PR #22, opened 2017-05-08
Add CMake files

uemoto1:
This pull request provides the CMake files for the binary builds based on (https://github.com/yhirokawa/SALMON/tree/cmake).
The usage of the build environment is written as below:

## Build

We use [CMake](https://cmake.org/) cross-platform build tools.

CMake detects the following configurations automatically,

- MPI Fortran/C compiler
- OpenMP compile flag
- LAPACK(/BLAS) and ScaLAPACK libraries

CMake software version **must** be 2.8 or later.
We recommend *3.0* or later versions.

### for your computer

    $ mkdir ./build_temp
    $ cd build_temp
    $ ../configure.py && make
    or
    $ mkdir build_temp
    $ cd build_temp
    $ cmake .. && make

You can specify compilers and/or compiler options.

    $ ../configure.py FC=mpiifort CC=mpiicc FFLAGS="-xAVX" CFLAGS="-restrict -xAVX"

### for Supercomputer systems and clusters

We provide the build configuration of the systems with CMake cross-compile mode.

    $ mkdir build_temp
    $ cd build_temp
    $ ../configure.py --arch=<COMPILER>-<SYSTEM> && make
    or
    $ mkdir build_temp
    $ cd build_temp
    $ cmake -D CMAKE_TOOLCHAIN_FILE=<COMPILER>-<SYSTEM> .. && make

For example, the following command builds the application for K-computer.

    $ ../configure.py --arch=fujitsu-k

CMake searches the cross-compile configuration files below `platform` directory.
If you want execution at the system that configuration file is not provided, you can create it yourself.

## Acknowledgement
- This PR is based on [Y.Hirokawa's patch](https://github.com/yhirokawa/SALMON/tree/cmake).
The author thank Mr Y. Hirokawa and Dr. S. Sato for fruitful discussions and comments.


shunsuke-sato:
@uemoto1 san, thank you for the development. I have confirmed that this patch gives the same results as before. Just, I want to ask one thing before merge.
Now, CMake build provides "SALMON.cpu", while Makefile build "salmon.cpu". There is no reason to have two different names for the different builds. So, just keep "salmon.cpu" for CMake build.

shunsuke-sato:
@uemoto1 san, thank you for the development. I have confirmed that this patch gives the same results as before. Just, I want to ask one thing before merge.
Now, CMake build provides "SALMON.cpu", while Makefile build "salmon.cpu". There is no reason to have two different names for the different builds. So, just keep "salmon.cpu" for CMake build.

uemoto1:
test this please

PR #28, opened 2017-05-24
Bug-fix; typo in the name of atomic species

shunsuke-sato:
I fixed the typo.


mn2007:
I think no one would complain about this modification.

PR #32, opened 2017-05-26
Optimized hpsi and Laplacian in GCEED.

shunsuke-sato:
I found a possibility of optimization for hpsi and Laplacian in the GCEED part.
The optimization is just based on factorization (see implementation).

Elapsed time for example `scf` is shortened from `67.91` sec. to `40.36` sec. on my environment. Furthermore, the elapsed time for CG method is shortened from `34.07` sec. to `10.95` sec.


mn2007:
Thank you for optimization.  I add this PR to whitelist.

PR #35, opened 2017-05-29
modify copyright in GCEED part

mn2007:
I modify copyright in GCEED part. (name of each developer -> SALMON developers)


uemoto1:
Thank you! I confirm this PR is okey to merge. (But take care to some confliction..)

shunsuke-sato:
@mn2007 san, and @uemoto1 san, I think no one complains this PR, and we do not need to wait for responses of other developers. So, could you fix this conflict and merge this PR to avoid further conflicts?

mn2007:
I resolved the conflicts but I'm not sure that person who should resolve the conflicts is the one who gave PR.   If my action was wrong, please tell me.

shunsuke-sato:
@mn2007 san, thank you for fixing the problem. Now, I could not manage time to see the code, but I'll review and merge it later, if anyone do not so at that time.
I personally think that owner of PR should fix conflicts if it is trivial since it is the easiest solution. But, I know it can be annoying... Quick review & merge would be only the solution to keep the developing environment comfortable...

mn2007:
I understand that it depends on cases. Thank you for your comment, @shunsuke-sato san.

PR #38, opened 2017-06-01
CMake: add USE_SCALAPACK option.

yhirokawa-ccs:
I added cmake build option: USE_SCALAPACK

In using CMake directly,

    $ cmake -D USE_SCALAPACK=yes ..

In using configure.py,

    $ ../configure.py --scalapack


mn2007:
Thank you for implementation of selection for eigen_subdiag_${FILE_MATHLIB}.f90.

PR #50, opened 2017-06-06
Name of pseudopotential files can be directly chosen.

shunsuke-sato:
I modified a way to choose pseudopotential files. Now, the name of ps file is directly given in input.


uemoto1 (approved):

mn2007:
I checked a GS calculation and it looks ok.

PR #52, opened 2017-06-07
Enabled the Fortran2003 standards.

yhirokawa-ccs:
I looked the following warning/remark messages.

    warning #5268: Extension to standard: The text exceeds right hand column allowed on the line.
    warning #5462: Global name too long,
    warning #6477: Fortran 2003 does not allow this statement or directive.
    warning #6843: A dummy argument with an explicit INTENT(OUT) declaration is not given an explicit value.
    warning #6916: Fortran 2003 does not allow this length specification.
    warning #7025: This directive is not standard F2003.
    warning #7346: The CHARACTER* form of a CHARACTER declaration is an obsolescent feature in Fortran 2003.
    warning #7416: Fortran 2003 does not allow this intrinsic procedure.
    remark #7712: This variable has not been used.

I enabled compiler checking for conforming the standards and fixed some compiler warnings from Intel compiler,

- fixed
    - 5268 (at ARTED/*.f90)
    - 6843 (at src/core/exc_cor.f90)
    - 6916, 7346 (at mpif.h of Intel MPI)
- suppressed
    - 6477, 7025

**For developers: I think that we should remove unused variables before sending pull requests.**


shunsuke-sato:
@yhirokawa san and developers, I agree that we should remove unused variables to keep the code clean!

PR #55, opened 2017-06-07
Fixed format for input dumping and enabled to check errors in input.

shunsuke-sato:
I fixed the format for input damping. And, I enabled to check errors in input and let SALMON stop if it finds the errors.


mn2007:
I add this PR to whitelist.

PR #60, opened 2017-06-09
Replaced communication module to salmon_communication in ARTED part.

yhirokawa-ccs:
I replaced ARTED/communication module to salmon_communication, and verified the results of both calculation modes.
I would be grateful if you could check the calculation results.


shunsuke-sato (commented):
@yhirokawa san, thank you for taking care of the communication in ARTED part! I have reviewed, and it's almost fine for me. But, let me add comments:
I think the Maxwell communicator should be replaced by the global one since it is also used for single-cell calculations without Maxwell eq. Furthermore, at the moment, we redundantly solve the Maxwell eq. and do not parallelize it. When we parallelize it, we will need the three communicators, tdks, maxwell, and global.
So, I would suggest that we should have the two communicators, tdks and global, for the moment.

yhirokawa-ccs:
@shunsuke-sato san, thank you for your review and PR! I merged it and rebased a branch for fixing conflicts.

PR #62, opened 2017-06-10
Fixed a bug in gceed part.

shunsuke-sato:
I fixed a bug in gceed part, which induces a compile error in gnu environment.


mn2007 (approved):
Thank you so much for fixing the bug ... .

PR #63, opened 2017-06-10
Implemented a new reader of atomic coordinates

shunsuke-sato:
- I implemented a function that reads the atomic coordinates. So far, GCEED part reads the atomic coordinates from a file in Cartesian coordinates, while ARTED part read that from input in reduced coordinates.

- I would say this is a bit confusing for users, and thus, I extended the function.
Now, one can give the atomic coordinates either from a file (`file_atom_red_coor/file_atom_coor`) or input  (`&atomic_red_coor/&atomic_coor`). Also, either reduced or Cartesian coordinates can be used.

- I removed results of GCEED/example since some of them can not be reproduced from example input anymore.


shunsuke-sato:
@uemoto1 san and @mn2007 san, I would like to ask you to review this PR. If both of you approve this, please merge this.

mn2007:
In previous meetings, there was a controversy whether atomic coordinates should be separated or not. GCEED handles large number of atoms, so coordinates are separated prevously. But come to think of it, I always write huge lines of atomic coordinates in input files when I use quantum espresso or cp2k. So, I think this modification causes no problem.

I have a question. We don't include output in examples even when we release stable version?

shunsuke-sato:
@mn2007 san, thank you for your comments. I did not know the discussion about the atomic coordinates, but I can imagine that the additional-file-input would be preferable for large systems. On the other hand, for standard users, the direct-input would be convenient. Thus, I extended SALMON to be able to employ either ways. So, if you want, still you can use an additional file `file_atom_coor` for atomic coordinate (see manual/input_variables.md for details).

About output, honestly, I do not see any reason why we have them in the code. They can be provided from input. Also, we can put them on web-tutorial. So, why do we need them in the code?

mn2007:
I checked some famous applications (cp2k and octopus) and found that there was no output in samples. I had thought that it's natural that samples contain output because RSDFT and SMASH have it. I understand your opinion and I change my mind. 

I'm glad if you add option that we can use seperated atomic coordinate file because GCEED developers are used to use such file, but this is not urgent task.

mn2007 (approved):
I approve this PR.

shunsuke-sato:
@mn2007 san, thank you for reviewing. 
Probably, my explanation was bad. Even at this moment, you can employ an additional file for atomic coordinates by using `&system/file_atom_coor` as before. For example, you can add `file_atom_coor='coo.dat'` in your input and remove `&atomic_coor`.

mn2007:
I understood what you wanted to say just after I commented. It was my fault. Thank you for your implementation.

uemoto1 (approved):

PR #65, opened 2017-06-11
change scalar variables for nstate and nelec to array 

mn2007:
I change scalar variables for nstate and nelec to array whose length is 2. This modification is needed because I have to eliminate previous input variables MST and ifMST from present input file. 

Since I apply this modification, some parts in ARTED are changed. Could you check carefully if this modification is correct or not?


shunsuke-sato:
@mn2007 san, thank you for the implementation.
However, I would a bit disagree this implementation. I'm sorry about that.
The problem is that, due to this implementation, the variable of "nelec"  starts having two different meanings: One is the total number of electrons. The other is the number of up-spin electrons. The mean of this variable now depends on another input variable, "ispin".
I would say that the multi-meaning with single-variable is quite confusing not only for users but also for developers, and potentially makes trivial bugs. Thus, let me suggest alternative solution: instead of modifying "nelec" and "nstate", one would introduce new input variables such as "nelec_spin(2)" and "nstate_spin(2)".
What do you think?

shunsuke-sato:
@mn2007 san, I implemented the alternative solution, which I suggested above, and sent its pull-request, PR #66. So, I would like to ask you to consider and review this solution.

mn2007:
As I commented on PR#66, I approve your PR. So, I close this PR. Some changes I'd like to pass are still remains. For these changes, I'll give another PR.

PR #66, opened 2017-06-11
Alternative solution of PR#65

shunsuke-sato:
I implemented an alternative solution of PR #65 in order to avoid multi-meaning with single-variable.


mn2007 (approved):
OK. I understand. Avoidance of confusion is quite important. Thank you for your implemetation.

PR #69, opened 2017-06-12
Clean up SALMON

shunsuke-sato:
- Wrapped routines for reading input.
- Dumping the input variables to `variables.log` instead of stdout since this information is not always necessary.
- To reduce warnings, I replaced `imag` with `aimag`, fitting Fortran 2003 standard.
- To reduce warnings, I removed a part of unused variables, routines and useless lines.


uemoto1 (approved):
Thanks, I confirm and approve this PR.

yhirokawa-ccs:
Thank you!
NOTE: `aimag` returns `real` with `kind` value. 
http://fortranwiki.org/fortran/show/aimag

PR #72, opened 2017-06-13
Implemented unit conversion for output.

shunsuke-sato:
- I added a header for the current-density file and implemented the unit conversion for the current. 

- I would say that this implementation is not the best solution, but a compromise between readability of output and convenience for maintenance/extension of the code.
If anyone has a good idea, please let us know. Otherwise, I'll continue to work based on this implementation.

### DIFF
--- a/ARTED/control/control_sc.f90
+++ b/ARTED/control/control_sc.f90
@@ -29,6 +29,7 @@ subroutine main
   use salmon_communication, only: comm_bcast, comm_sync_all, comm_is_root
   use misc_routines, only: get_wtime
   use salmon_global, only: format3d, out_dns, out_dns_rt, out_dns_rt_step
+  use inputoutput, only: t_unit_time, t_unit_current, t_unit_ac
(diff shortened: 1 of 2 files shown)


mn2007:
Thank you for your implemetentation. I might use this kind of modification for polarizability or something. Could you continue to implement?

shunsuke-sato:
Thank you for your comment. Here, I tried to prepare the unit-conversion variables at beginning of the calculation, and to avoid computing on the fly in order to avoid the following situation: developers have to revisit all over the codes when the unit part is modified.

OK! I'll continue to work! But, could you merge this PR so that other developers could kindly participate the output-file modification.

mn2007:
I'd like to merge if I can. But the modification affects to the ARTED part. So, I'd like to wait other people comment. Sorry for that.

uemoto1 (approved):
Thanks Sato-san, Okey, I confirm this PR.

PR #73, opened 2017-06-14
Add communication routines and variables.

yhirokawa-ccs:


mn2007:
I cannot follow for detail clearly but this modification includes isend and irecv, and that part is done for the GCEED. Thank you. I agree this modification.

mn2007:
I understand this modification. This is clearly done for GCEED. After all checks pass, I merge PR.

PR #74, opened 2017-06-14
Wrapped all communication routines in GCEED part.

yhirokawa-ccs:
I wrapped all communication routines in GCEED part.

I verified the calculation results with C2H2 molecule samples (sample 1-3), but this PR may exists any bugs because this is a big modification.
Would you please check the calculation results under GCEED part?

### DIFF
--- a/GCEED/common/OUT_IN_data.f90
+++ b/GCEED/common/OUT_IN_data.f90
@@ -17,8 +17,7 @@
 
 SUBROUTINE OUT_data
 use salmon_parallel, only: nproc_id_global, nproc_size_global, nproc_group_global, nproc_group_h
-use salmon_communication, only: comm_is_root
-use mpi, only: mpi_integer, mpi_sum, mpi_double_precision
+use salmon_communication, only: comm_is_root, comm_summation, comm_bcast
 use scf_data
 use new_world_sub
 use read_pslfile_sub
@@ -41,7 +40,6 @@ integer :: nproc_xyz_datafile(3)
 character(8) :: fileNumber_data
 integer :: iob_myob
 integer :: icorr_p
-integer :: ierr
 
 if(comm_is_root(nproc_id_global))then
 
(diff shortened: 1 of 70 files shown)


mn2007:
I understand. I'll try to do tests under some conditions. 
Maybe I'll make a responce to this tomorrow morning.

mn2007:
I'm trying to check and I found that it may take time to implement this PR. 
Sorry for that and I appreciate your contribution.

First comment is about deadlock. What I did was to check GS calculation with FX10 12 nodes modifying the input from 
<pre>
  nproc_domain = 1,1,1
  nproc_domain_s = 1,1,1
</pre>
to
<pre>
  nproc_domain = 3,4,1
  nproc_domain_s = 3,4,1
</pre>
As a result, the calculation stopped because of dead lock of communication. A part where the calculation stopped was around line 250 in GCEED/common/hartree_boundary.f90. GCEED used variable "nproc" as the number of total processors, not as rank. So, when total number of processors is 1, copy of array is done in that part, and when the total number of processors is larger than 1, allreduce is done. 

I guess that easiest way to solve the problem is to replace from "nproc_id_global==1" to "nproc_size_global=1". Using the grep command, I found that there are 6 parts where it is written "nproc_id_global==1". Could you apply this modification at first?

yhirokawa-ccs:
@mn2007 san, thank you for reporting! I solved a deadlock with your suggestion, but some bugs still exist. Maybe I will send a bugfix commit to next week.

mn2007:
@yhirokawa san, thank you for your reply. It's really tough work for me, so I prefer to do this kind of work next week.

yhirokawa-ccs:
I sent a pull request #76 for solving bugs. Would you please check and merge it?

mn2007:
As I mentioned PR#76, I merged it. Thank you.

yhirokawa-ccs:
@mn2007 san, thank you for merging. I rebased it because Sato-san fixed freeze bug in GCEED part.
I verified the results under knights landing 16 nodes.

    nproc_domain = 4,2,2
    nproc_domain_s = 4,2,2

Would you please check the results?. But there is not rush, so please take your time.

mn2007:
I understand. I'll do some tests on Monday. Although I didn't tell you that there was parts that a process sending data in bcast is not rank 0, you dealed with that. I really appreciate your efforts.

mn2007:
All I have to do is to figure out why ground state calculation stops at the end when following condition is specified for the calculation using FX10 12node. 
<pre>
  nproc_ob=2
  nproc_domain = 3,2,1
  nproc_domain_s = 3,4,1
</pre>
All other checks have passed. After I figure out the reason, I'll inform you.

mn2007:
The calculation stops even if I use other branch, so it's my fault. Could you wait until I figure out the reason. Thank you for your patience.

mn2007:
Since it's my fault, I'll give another PR that fix this problem. I merge this PR.

mn2007 (approved):
I approve this PR.

PR #77, opened 2017-06-15
Prepared explicit choice for the BJ exchange.

shunsuke-sato:
- For usability, I prepared an explicit  choice for the Becke-Johnson potential with PW correlation, `BJ_PW`.
Originally, this choice was implicitly possible with `functional=TBmBJ` and `cval=1.0`. However, this way of the selection seems a bit weird. 

- I also changed the default value for the c-parameter of TBmBJ potential to `cval=-1.0`, negative value, which triggers the internal evaluation of the parameter based on the expression in the original paper.

### DIFF
--- a/ARTED/RT/dt_evolve.f90
+++ b/ARTED/RT/dt_evolve.f90
@@ -156,7 +156,7 @@ Subroutine dt_evolve_omp_KB(zu)
 
 ! yabana
   select case(functional)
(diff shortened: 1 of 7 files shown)


uemoto1 (approved):
Thank you, @shunsuke-sato san!
I confirmed your change, and understood we can continuously use "TBmBJ, cval=1" for compatibility.
I think your PR is okey to merge!

PR #78, opened 2017-06-16
Hotfix: Added missing broadcast.

shunsuke-sato:
@yhirokawa san, I fixed a bug in `modules/inputoutput.f90` and confirmed that SALMON runs properly even in parallel environment. Could you check it on your environment, and merge this PR if it passes.


yhirokawa-ccs:
@shunsuke-sato san, thank you for quick fix!
I made sure that the calculation is successfully finished. I will merge it.

PR #81, opened 2017-06-22
fix a bug for orbital parallelization in the GCEED part

mn2007:
There was a bug in orbital parallelization for a long time. So, I fix it. We can obtain same results even if we specify
<pre>
  nproc_ob = 2
  nproc_domain = 3,2,1
  nproc_domain_s = 3,4,1
</pre>
 
In this bug fix, I also modify salmon_communication.f90. Could you examine whether it's acceptable or not, @yhirokawa san? If it's acceptable, you can merge this PR without any comment.


yhirokawa-ccs:
@mn2007 san, I apologize for the delayed response.
It feels like no problem. I will merge it.

PR #82, opened 2017-06-23
make automatically calculate imesh_oddeven

mn2007:
By the automated calculation for the variable "imesh_oddeven", I remove the variable from read_input_scf.f90, and avoid to pick up the variable from read_input_scf.f90 to inputoutput.f90.

### DIFF
--- a/GCEED/common/CMakeLists.txt
+++ b/GCEED/common/CMakeLists.txt
@@ -38,6 +38,7 @@ set(SOURCES
     setcN.f90
     set_filename.f90
     set_gridcoo.f90
+    set_imesh_oddeven.f90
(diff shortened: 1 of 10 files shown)


shunsuke-sato:
@mn2007 san, thank for the implementation. I tested this patch with the example inputs. It works for the ground state, but it stops when we run RT mode. SALMON emits the following message:
`You cannot use old data files that values of Lsize/Hgs are a mixture of odd and even.`

Could you check the example input?

mn2007:
Thank you for your check, @shunsuke-sato san. I modify the part that calculation stops at. Sorry to have bothered you.

PR #83, opened 2017-06-24
Implemented DoS writing function

shunsuke-sato:
First, I moved output variables in `sys_name_DoS.out` into `sys_name_band.out`.
Then, I implemented a function to write the density of states in  `sys_name_DoS.out`. Thus,  `sys_name_DoS.out` contains different data from previous versions.

Now, SALMON computes the DoS for solid state systems with Lorentzian broadening (0.1 eV width) at the end of GS calculation, if `&analysis/out_dos` is set to `'y'`.


uemoto1:
Thank you Sato san.
I think it is okey to merge.

PR #85, opened 2017-06-24
Decomposed main programs of ARTED part

shunsuke-sato:
- I separated initialization (in `initialization.f90`) and ground state calculation (in `ground_state.f90`) from main programs of ARTED part (`control_sc.f90` and `control_ms.f90`) because the main programs were too long and contained too many functions. Also, the two main programs had a big overlap.

- This patch is the preparation for (i) separation of  `GS` and `RT` calculation modes, and (ii) restart from backup.


uemoto1 (approved):
Okey, I confirmed the modificated part of initialization.
Thank you for developing! @shunsuke-sato san.

PR #88, opened 2017-06-25
Fix a problem with OpenMP in GCEED part

shunsuke-sato:
If I understand correctly, loop counters of inner loops are not guaranteed to be `private` with the OMP parallelization. Thus, I added explicit `private` statement accordingly.

Probably, my statement is not correct. It seems that the inner loop counters are implicitly declared as `private` by Fortran default. But, nevertheless, I would say that the explicit declaration would be desirable.


mn2007:
Thank you for your suggestion, @shunsuke-sato san. I examined whether it's necessary or not, and found that the variable in inner do loop becomes shared. The explanation is given in a following lecture material. 
http://www.aics.riken.jp/aicssite/wp-content/uploads/2017/04/tokuronA_17_3_katagiri_r1.pdf
(around p.49. sorry for giving a japanese one.) 

So, I accept your proposal.

mn2007 (approved):
I approve this PR.

PR #91, opened 2017-06-28
use sup in the developer's manual

mn2007:
@shunsuke-sato san used "sup" in the manual. So, I also use it.


uemoto1:
Thank you, @mn2007. I will merge your contribution!

PR #92, opened 2017-06-28
implement the Broyden's method

mn2007:
I implement the Broyden's method in the GCEED part. For isolated systems, we can choose simple mixing or Broyden's method by setting the variable "amixing". 

I can't figure out who should check this PR. Could someone check?

### DIFF
--- a/GCEED/modules/scf_data.f90
+++ b/GCEED/modules/scf_data.f90
@@ -123,8 +123,6 @@ integer :: max_lg_num
 integer :: Miter       ! Total number of Iteration for SCF calculation
 integer :: Miter_rt    ! Total number of Iteration for RT calculation
 
-real(8) :: mixrate     ! Mixing rate for updating charge density
-
 integer :: iflag_diisjump
 real(8) :: lambda1_diis, lambda2_diis
 
--- a/GCEED/scf/CMakeLists.txt
+++ b/GCEED/scf/CMakeLists.txt
@@ -1,4 +1,6 @@
 set(SOURCES
+    add_in_broyden.f90
+    broyden.f90
     calc_dos.f90
     calc_occupation.f90
     calc_pdos.f90
--- a/GCEED/scf/add_in_broyden.f90
+++ b/GCEED/scf/add_in_broyden.f90
@@ -1,0 +1,33 @@
+subroutine add_in_broyden(trho,trho_in,tfb,trho_temp)
+  use inputoutput
+  use salmon_parallel, only: nproc_group_orbital
+  use salmon_communication, only: comm_summation
+  use scf_data
+  implicit none
+  integer :: ix,iy,iz
+  integer :: nn,nn2
+  real(8) :: trho(mg_sta(1):mg_end(1),mg_sta(2):mg_end(2),mg_sta(3):mg_end(3))
+  real(8) :: trho_in(mg_sta(1):mg_end(1),mg_sta(2):mg_end(2),mg_sta(3):mg_end(3))
+  real(8) :: tfb(mg_sta(1):mg_end(1),mg_sta(2):mg_end(2),mg_sta(3):mg_end(3))
+  real(8) :: trho_temp(mg_sta(1):mg_end(1),mg_sta(2):mg_end(2),mg_sta(3):mg_end(3))
+  
+  trho(:,:,:)=trho_in(:,:,:)+alpha_mb*tfb(:,:,:)-trho_temp(:,:,:)
+  
+  nn=0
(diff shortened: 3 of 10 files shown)


shunsuke-sato:
@mn2007 san, thank you for your implementation. If no one reviews this patch, I'll take care of it. But, most probably, this weekend.
I had a quick look at the implementation, and I have a couple of comments/questions.
1. Do you have any evidence to show that the implementation is correct? If you clearly show a typical result of a new function, probably, the review process would become easier.
2. Are there any reasons why you implemented the matrix inversion by yourself? I would suggest to use the Lapack routine since it would be faster and more accurate, and less bugs.

mn2007:
Thank you for your review, @shunsuke-sato san. I answer your questions.

1. I've checked the sample of C2H2. With simple mixing, I obtain a following result
<pre>
 iter =   101     Total Energy =      -339.70410465     Vh iteration =  21
     1       -18.4493      2       -13.9885      3       -12.3937      4        -7.3311
     5        -7.3311
iter and ||rho(i)-rho(i-1)||**2/(# of grids) =    101 0.55535234E-16
</pre>
and with the Broyden's method, I obtain a following result. 
<pre>
 iter =    47     Total Energy =      -339.70413808     Vh iteration =   2
     1       -18.4492      2       -13.9884      3       -12.3936      4        -7.3310
     5        -7.3310
iter and ||rho(i)-rho(i-1)||**2/(# of grids) =     47 0.20969326E-16
</pre>
We can see that same results are obtained and an iteration of Broydens' method is less than that of simple mixing. Dipole moments in the z-direction at 5000 rt loops in an optical response calcluation are -0.10486778E-02 for simple mixing and -0.10425003E-02 for Broyden's method. We can say they match.  

2. This implementation is grid parallelized version of the Broyden's method in the ARTED part, and matrix inversion is completely same as yours. I think that there is a rule that routines in the ARTED/CCEED parts can not call subroutine each other unless they are in core libraries. Using lapack might be future work, but we obtain same result as the one of simple mixing at this moment, and I think it is not urgent.

mn2007:
@shunsuke-sato san, I implement other part if you check this weekend, so there is no need to hurry.

shunsuke-sato:
@mn2007 san, thank you for your quick response! It loos fine! 
For the matrix inversion, let me explain the situation in ARTED part. ARTED originally used home-made routines for linear algebra such as diagonalization. At some point, I switched to use the Lapack routines, but some parts still remain to use the home-made routines. But, this does not mean the recommendation to provide the home-made routine. Furthermore, I would also not recommend the duplication of the same code since it just increases the management cost...
Anyway, I'll replace the matrix inversion by Lapack routine later.
Thanks for the implementation.

mn2007:
Please let me comment on the dupurication. If we admit to call routines each other, we must continue to check whether the routines exist on the other side because the modifications of the routines affect to the other side. In addition, since we use different ways of parallelization, it is not garanteed that routines which have same functions become completely same. Maybe we must put them in core libraries by mutual agreeement to avoid the duplication and I think we need further discussion.  

I wait if you replace the matrix inversion by Lapack routine. So, please do not hurry.

shunsuke-sato:
I understand your point, but, therefore, I suggested to use Lapack, which avoids all the problems you raised in this case. And, Lapack is already linked to SALMON. So, we can just call it!

Also, I would say that we must continue to check whether the same routine exists somewhere else in SALMON. If it exists, we should use or extend it, and might move it into core. Otherwise, we would have a lot of similar routines, which are called slightly differently, and make the code difficult to manage in the end.

shunsuke-sato:
@mn2007 san, thank you for the implementation. I have reviewed it, and it is fine. Before merging it, let me ask one question. Why do you add the following sentence in the variable explanation? Are there any problems?
`If <code>&system/iperiodic</code> is <code>0</code>, <code>nmemory_mb</code> must be less than 21.`

mn2007:
I answer your question. We stock rho_in and rho_out for 20 loops before the scf iteration at that time. This reduces not only the amount of allocated memory but also reading time in reentrance of the scf calculation. 
I had noticed that your side stocks rho_in and rho_out for all scf iterations, and it may cause a shortage of memory in our case. So I'm afraid that we need negotiations when we prepare core routines. Anyway, we can talk them in Skype meeting.

shunsuke-sato:
@mn2007 san, thank you for the comment. I understand the point, and I think ARTED part also should not have all the information. And, probably, we should modify the code so that it only stores the data for `nmemory_mb` scf iterations in the near future.

PR #93, opened 2017-06-29
make calculate num_pole_xyz automatically

mn2007:
I make calculate num_pole_xyz automatically. Calculated num_pole_xyz become the ones as in previous input. So, results do not change for C2H2.

### DIFF
--- a/GCEED/scf/read_input_scf.f90
+++ b/GCEED/scf/read_input_scf.f90
@@ -274,7 +274,7 @@ else if(meo==3)then
     stop "num_pole_xyz must be largar than 0 when they are not default values."
   end if
   if(num_pole_xyz(1)==0.and.num_pole_xyz(2)==0.and.num_pole_xyz(3)==0)then
-    dip_spacing = 8.d0*au_length_aa/ulength_from_au  ! approximate spacing of multipoles 
+    dip_spacing = 8.d0/au_length_aa  ! approximate spacing of multipoles 
     num_pole_xyz(:)=int((al(:)+dip_spacing)/dip_spacing-1.d-8)
   end if
 end if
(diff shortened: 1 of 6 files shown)


GCEED/scf/read_input_scf.f90
```diff
@@ -256,26 +257,28 @@ subroutine read_input_scf(iDiterYBCG,file_atoms_coo)
...
+if(meo<=0.or.meo>=4)then
+  stop "meo must be equal to 1 or 2 or 3."
+else if(meo==3)then
+  if(num_pole_xyz(1)==0.and.num_pole_xyz(2)==0.and.num_pole_xyz(3)==0)then
+    continue
+  else if(num_pole_xyz(1)<=0.or.num_pole_xyz(2)<=0.or.num_pole_xyz(3)<=0)then
+    stop "num_pole_xyz must be largar than 0 when they are not default values."
+  end if
+  if(num_pole_xyz(1)==0.and.num_pole_xyz(2)==0.and.num_pole_xyz(3)==0)then
+    dip_spacing = 8.d0*au_length_aa/ulength_from_au  ! approximate spacing of multipoles
```
shunsuke-sato: Here, `dip_spacing` and `num_pole_xyz` seem to depend on the input unit since `ulength_from_au` depends on the input unit. However, I think these values should not have the unit dependency. Otherwise, the results can depend on the unit even if one employs the physically same parameters.

shunsuke-sato (commented):
@mn2007 san, thank you for the development. I reviewed it, and wrote a comment here. So, please check it.

mn2007:
It' easier for me to reply this comment than others. So I reply to this comment first.
You are right. Variables except for input and output should be in atomic unit. So I modify the calculation of dip_spacing. Thank you for your review, @shunsuke-sato san.

shunsuke-sato:
@mn2007 san, thank you for your quick correction! I'll continue review and merge it!

PR #94, opened 2017-06-29
add the variable numfiles_out_3d

mn2007:
I add the variable "numfiles_out_3d" in inputoutput.f90 instead of picking up the variable "numfile_movie" in read_input_rt.f90. Using FX10 12nodes, I checked the GS calculation with following condition
<pre>
&parallel (<- parallel, somehow it is not appropriately displayed)
  domain_parallel = 'y'
  nproc_ob = 1
  nproc_domain = 3,4,1
  nproc_domain_s = 3,4,1
/
...
&analysis
  out_psi = 'y'
  format3d = 'avs'
  numfiles_out_3d = 2
/
</pre>
and confirmed that it certainly produces two files for each psi. I used AVS/express combining files for psi and confirmed that HOMO was displayed. 

This PR is what I planned to give tomorrow, so I will not give other PRs tomorrow.

### DIFF
--- a/GCEED/common/CMakeLists.txt
+++ b/GCEED/common/CMakeLists.txt
@@ -39,6 +39,7 @@ set(SOURCES
     setcN.f90
     set_filename.f90
     set_gridcoo.f90
+    set_icoo1d.f90
     set_imesh_oddeven.f90
     set_ispin.f90
     set_isstaend.f90
--- a/GCEED/common/OUT_IN_data.f90
+++ b/GCEED/common/OUT_IN_data.f90
@@ -645,7 +645,10 @@ else if(ilsda==1)then
   end if
 end if
 
-if(iSCFRT==2) call allocate_mat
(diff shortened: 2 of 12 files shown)


mn2007:
The reason why I prepared this variables is to use in the K computer. The K computer has only 14 GB in each node. So if the master genarates AVS files in some rt time steps, the disk amount for these files will be more than 14 GB. In addition, parallelization for writing files reduces elapsed time. Combining separated files can be done by cat command in Linux.  However, I'm afraid that the person who uses this function is only me.

shunsuke-sato:
@mn2007 san, thank you for your implementation. I think the parallelization of I/O would be useful, and thus, I think this PR is worth to be merged. But, I did not merge it because I'm not familiar with these out put formats, and thus, I expected that someone reviews it.
But, if no one reviews this, I'll do it tonight.

PR #96, opened 2017-07-01
Fixed a segmentation fault in tddft_sc.

yhirokawa-ccs:
I fixed uninitialized variable in tddft_sc.

### DIFF
--- a/ARTED/control/control_sc.f90
+++ b/ARTED/control/control_sc.f90
@@ -57,6 +57,7 @@ subroutine tddft_sc
 !====RT calculation============================
 
   call init_Ac
+  iter=Nscf+1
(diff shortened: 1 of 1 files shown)


shunsuke-sato:
@yhirokawa san, thank you for fixing the segmentation fault, which is induced by my PR. In the original code, `iter` is equal to `Nscf+1` here, as you fixed. 
However, I found that it was not correct, and `iter` should be zero here. 
I sent an PR to your branch. So, could you merge it, please? Then, I can merge this PR.

shunsuke-sato:
@yhirokawa san, thanks a lot for your quick response! I'll merge it.

PR #97, opened 2017-07-01
Implemented 'GS' and 'RT' modes in addition to 'GS_RT' mode in ARTED part

shunsuke-sato:
To reduce functional difference between ARTED and GCEED, I implemented new calculation modes (`calc_mode`): `GS`  and `RT` in addition to `GS_RT`.

- `GS` mode computes the ground state and writes the ground state wavefunctions.
- `RT` mode reads the ground state wavefunctions, which are computed in `GS` mode, and then computes the electron dynamics, 
- Now, the multi-scale simulation with `RT` mode can start based on the ground state wavefunctions from the single-cell calculation with `GS` mode.
- In the present implementation, `GS` mode writes the ground state wavefunction in `&control/directory`. Because the number of the files can be huge (as many as k-points), we should prepare an additional directory for the wavefunction data, at the latest, until the first official release. A simple solution would be to allow to use `CUTE_COMMAND_LINE`, which is supported by Fortran2008 and later.

### DIFF
--- a/ARTED/GS/CMakeLists.txt
+++ b/ARTED/GS/CMakeLists.txt
@@ -1,4 +1,5 @@
 set(SOURCES
+    io_gs_wfn_k.f90
     Gram_Schmidt.f90
     diag.f90
     sp_energy.f90
--- a/ARTED/GS/io_gs_wfn_k.f90
+++ b/ARTED/GS/io_gs_wfn_k.f90
@@ -1,0 +1,106 @@
+!
+!  Copyright 2017 SALMON developers
+!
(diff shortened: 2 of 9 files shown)


shunsuke-sato:
Are there anyone will review this PR? If not, I'll merge it by myself. Then, on top of this PR, I will reconstruct the backup routine of ARTED part this weekend.

uemoto1:
Thank you @shunsuke-sato san for development. (Sorry for late reply..)
I have confirmed the modification, which makes possible to select `RT` and `GS` mode with exporting the GS state to the files. I agree this is okey to merge to develop branch.

PR #98, opened 2017-07-07
Implemented error function and its complement.

shunsuke-sato:
I implemented the error function `erf_salmon(x)` and its complement `erfc_salmon(x)` based on the following reference: W. J. Cody Math. Comp. 23 631 (1969).

Now, SALMON meets the Fortran2003 standard.

### DIFF
--- a/ARTED/common/Exc_Cor.f90
+++ b/ARTED/common/Exc_Cor.f90
@@ -17,6 +17,7 @@
 subroutine Exc_Cor(GS_RT,NBtmp,zu)
   use Global_Variables
   use timer
+  use salmon_math
   implicit none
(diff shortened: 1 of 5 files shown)


uemoto1 (approved):
Thank you for implementation. 
I have checked that this code can reproduce the total energy without any significant differences, and I think it is okey to merge.
(Recommendation: I think it is helpful for the future readers to add a comment-line which indicates the reference to the original publication. )

shunsuke-sato:
@uemoto1 san, thank you for your review. Just, a comment. I have already added the reference in the code.

uemoto1:
Okey, I noticed a comment-line in modules/salmon_math.f90, sorry for mistake.

PR #99, opened 2017-07-08
Modified external field

shunsuke-sato:
I modified external fields.

- Implemented a new envelope `Ecos2` for the ARTED part.
- Merged the namelist of `linear_response` into `em_field`.

### DIFF
--- a/ARTED/RT/init_Ac.f90
+++ b/ARTED/RT/init_Ac.f90
@@ -78,8 +78,8 @@ Subroutine init_Ac
       end if
     enddo
   case('Ecos2')
-    if(phi_CEP1 /= -0.5d0)then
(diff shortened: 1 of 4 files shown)


mn2007:
Thank you for your implementation, @shunsuke-sato san. It's fine for me. (I've checked that a result for the GCEED part doesn't change.) @uemoto1 san, could you check this PR and merge?

mn2007:
I'm going to modify an example file for the GCEED part after this PR passes. This is just an announcement.

uemoto1:
Okey, I confirmed that the group of the impulse amplitude in the inputfile and the preferred CEP value of Ecos2 pulse are changed, and the code still reproduce the results obtained in previous version.

PR #101, opened 2017-07-10
Move fbessel_j1 to salmon_math

uemoto1:
This PR starts to move the defination of special function (j1), which is not a standered in F2003, to `salmon_math` module.  The algorithm used in the code is based on the power expansion is very primitive, and hard to refer the appropriate articles in source code. 

TODO: Add the reference when we impliments the improved algorith of the special functions for future.

### DIFF
--- a/ARTED/FDTD/beam.f90
+++ b/ARTED/FDTD/beam.f90
@@ -15,24 +15,6 @@
 !
 
 !===============================================================
-! NOTE: bessel_j1 is Fortran 2008 standard intrinsic procedure
-real(8) function fbessel_j1(x)
-  implicit none
-  real(8), intent(in) :: x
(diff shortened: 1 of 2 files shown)


shunsuke-sato:
@uemoto1 san, thank you for the implementation. I have reviewed it and will merge it. But, the present implementation is valid only for relatively small `x`, and thus, we need a comprehensive implementation to avoid wrong use of this function. I'll add it on an issue list.

uemoto1:
Thank you, @shunsuke-sato san.
I agree this definition is incomplete.
(However, for now, the salmon only employ this function only the in case of n=1 and small x).
I have an experience to implement the improved algorithm containing the asymptotic and iterative form based on numerical recipes. I can give it for near future.

PR #102, opened 2017-07-10
Support OpenMP only execution.

yhirokawa-ccs:
We will support OpenMP only parallelization (single node execution without MPI).
By this PR, CMake builds SALMON without MPI parallelization in default configuration, but architecture specific build (`--arch` option) enables MPI as usual.

I tested gcc 4.8.5 and icc 17.0.4 under Xeon Phi.
Could you please check build and calculation results?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,7 @@ include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/setup.cmake)
 project(SALMON Fortran C)
 
 # optimization options
+option_set(USE_MPI                        "Use MPI parallelization"                                  OFF)
 option_set(USE_TLOG                       "Use TLOG library"                                         OFF)
 option_set(USE_PAPI                       "Use Performance API"                                      OFF)
 option_set(USE_NVTX                       "Use NVIDIA Tools Extension Library"                       OFF)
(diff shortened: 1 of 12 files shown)


mn2007:
I'll check whether the code for GCEED part runs on FX10. I cannot clearly declare when I do this work, but I'll do it by this weekend. After that, I'll comment the results. If no one doesn't comment by that time, I'll merge this PR.

uemoto1:
@yhirokawa -san. Thank you for developing the single process environment. 
I will check in the `Intel(R) Xeon Phi(TM) CPU 7250` single node, and it works successfully.
Sometimes, I have failed to `make` with the parallel build in `arch=intel-knl` environment.
(The re-building will be success.) I suspect the current version has any problem in dependency chains of `CMakeList.txt`. 
```
[  5%] Building Fortran object modules/CMakeFiles/salmon_module.dir/salmon_parallel.f90.o
[  6%] [  6%] Building Fortran object modules/CMakeFiles/salmon_module.dir/inputoutput.f90.o
Building Fortran object modules/CMakeFiles/salmon_module.dir/salmon_math.f90.o
Linking Fortran static library libsalmon_module.a
[  6%] Built target salmon_module
```
Since this is not a major problem,  please check them when it has a time enough.

I will check other environments which I can use. Please wait.

mn2007:
Thank you for implementing, @yhirokawa san. And thank you for commenting, @uemoto1 san. I've confirmed that it successfully runs on FX 12 nodes. All I can do is up to this. @uemoto1 san, could you manage this PR?

uemoto1:
Okey, it seems to be okey in other intel compiler environment. I think it is possible to merge.
Thank you, @yhirokawa -san.

yhirokawa-ccs:
Thank you for comment and merge, @mn2007 san and @uemoto1 san.
@uemoto1 san, I fixed a parallel build error with PR #107. Could you please check it?

PR #105, opened 2017-07-13
modify values for ae_shape

mn2007:
I modify values for ae_shape in the GCEED part. I intend to implement the code based on the SALMON meeting, but there might be mistakes. In addition, I cannot modify the developer's manual because there are a lot of values in the ARTED part. 

I'd like to ask @shunsuke-sato san to review this PR. Could you please review? And is it possible for you to modify the manual after this PR passes?

### DIFF
--- a/GCEED/rt/calc_env_trigon.f90
+++ b/GCEED/rt/calc_env_trigon.f90
@@ -47,9 +47,9 @@ subroutine calc_env_trigon(ipulse,tenv_trigon)
     theta2=dble(itt)*dt-t1_t2-0.5d0*pulse_tw2+phi_cep2*2d0*pi
   end if
 
-  if(tae_shape=='ecos2')then
+  if(tae_shape=='Ecos2')then
     tenv_trigon=cos(alpha1*theta1)**2*cos(alpha2*theta2)
-  else if(tae_shape=='acos2')then
+  else if(tae_shape=='Acos2')then
     tenv_trigon=-(-alpha1*sin(2.d0*alpha1*theta1)*cos(alpha2*theta2)   &
                   -alpha2*cos(alpha1*theta1)**2*sin(alpha2*theta2))
   end if
--- a/GCEED/rt/calc_env_trigon.f90
+++ b/GCEED/rt/calc_env_trigon.f90
@@ -25,33 +25,33 @@ subroutine calc_env_trigon(ipulse,tenv_trigon)
   integer,intent(in)  :: ipulse  ! 1: first pulse, 2: second pulse
   real(8),intent(out) :: tenv_trigon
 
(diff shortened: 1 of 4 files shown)


shunsuke-sato:
@mn2007 san, thank you for the implementation. I reviewed it, and have a question. Why do you use a small letter `a/e` for the pulse shape? I think a capital letter would be more natural since `A/E` is the conventional notation for vector potential/ electric field. Or, do you forbid to use capital letters even in input?

mn2007:
Thank you for reviewing, @shunsuke-sato san. This was my mistake and I totally agree with you. I think we should actively use large capital letters for values of character variable depending on the situation.

GCEED/rt/calc_env_trigon.f90
```diff
@@ -31,27 +31,27 @@ subroutine calc_env_trigon(ipulse,tenv_trigon)
...
     tae_shape=ae_shape2
-    ! sin(alpha1*theta1)**2
+    ! cos(alpha1*theta1)**2
     alpha1=Pi/pulse_tw2
-    theta1=dble(itt)*dt-t1_t2
-    ! sin(alpha2*theta2) or cos(alhpa2*theta2)
+    theta1=dble(itt)*dt-t1_t2-0.5d0*pulse_tw2
+    ! cos(alhpa2*theta2)
     alpha2=omega2
     theta2=dble(itt)*dt-t1_t2-0.5d0*pulse_tw2+phi_cep2*2d0*pi
```
shunsuke-sato: @mn2007 san, thank you for the implementation. But, it seems that the definition of CEP does not much the conventional one, `cos(w*t+phi)`, since, in your implementation, the CEP depends on the frequency. I mean the argument of the carrier wave contains a product of `alpha2` and `phi_cep2*2d0*pi` in your implementation. However, the CEP part should be separated from the frequency in the conventional definition.

mn2007:
Thank you for your early reply. Since these three days are holidays, I'll consider and reply your review next Tuesday. Sorry for inconvenience.

GCEED/rt/calc_env_trigon.f90
mn2007: Thank you for reviewing, @shunsuke-sato san. It was my critical mistake. I modified the code according to your suggestion, and found that results change before and after the modification. I think modified one is correct.

shunsuke-sato:
@mn2007 san, thank you for the modification. It looks fine, but I'm not sure if you finished the implementation. Thus, I cannot merge it.
Did you finish the implementation, and can I merge this?

mn2007:
@shunsuke-sato san, thank you for saying it's fine and confirming the situation. I finished the implementation. So could you merge this?

PR #106, opened 2017-07-18
make calculate nproc_ob etc automatically

mn2007:
I think most people are being busy. But I think the pass of this PR makes all users happy. 
So could someone please merge this PR?

This modification enables users to run the code without any specification of namelist "parallel". 
If users use default values and use number of processers with the combination of prime factors including 2, 3 and 5, the code for the GCEED part automatically assigns nproc_ob, nproc_domain, and nproc_domain_s.  

I've checked for the GCEED part on FX10, and confirmed that results with 12, 144, and 160 processors
are not different.

At the moment, even if number of processes is too large compared with the system size, the calculation continues to run with giving wrong results. Maybe I'll modify this by next release of SALMON. In addition, I'll modify the manual too.

### DIFF
--- a/GCEED/rt/CMakeLists.txt
+++ b/GCEED/rt/CMakeLists.txt
@@ -12,6 +12,7 @@ set(SOURCES
     read_input_rt.f90
     read_rt.f90
     real_time_dft.f90
+    set_numcpu_rt.f90
     taylor_coe.f90
     taylor.f90
     time_evolution_step.f90
--- a/GCEED/rt/read_input_rt.f90
+++ b/GCEED/rt/read_input_rt.f90
@@ -97,9 +97,6 @@ if(iwrite_projection==1.and.itwproj==-1)then
 end if
 
 !===== namelist for group_parallel =====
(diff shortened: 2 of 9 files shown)


uemoto1:
Thank you mn2006 san for development.

I checked not affect the calculation in ARTED part and GCEED still works with removing `parallel` group.

I confirmed the automatic estimation of the process numbers (`set_numcpu_(scf|rt)`).
According to the previous meeting, we have a plan to prepare the two types of examples for the single and multi-processing environments. However, after this PR, we can use common input files for the both of environments, right?

I think it is helpful for the tutorials.

mn2007:
Thank you for checking, @uemoto1 san. Yes, you are right. After the pass of this PR, we need not prepare both documents for single and multi-process environments, and need not prepare both tutorials (samples?) for single and multi-process environments.

uemoto1:
Okey, I understand, Thanks @mn2007 san.

mn2007:
@uemoto1 san, could you please merge this PR? Then I can move on.

PR #107, opened 2017-07-19
Fixed a parallel build error is caused by incorrect dependencies.

yhirokawa-ccs:
I fixed an error of parallel build. It is caused by incorrect build dependencies.
I set all libraries depend a SALMON module library.

Could you please check parallel build?

    $ ../configure.py
    $ make -j8


uemoto1 (approved):

uemoto1:
Thank you, @yhirokawa -san.

I have successfully compiled the binary with increasing the number of parallelization `make -j 4` 
in my environment, and it can reproduce the previous result.
And, I confirmed the dirname of ARTED-related static library is renamed to `BUILD/ARTED/libarted_????.a`

I think this modification is very helpful and possible to merge.

shunsuke-sato:
@yhirokawa san, thank you for fixing the parallel build, and @uemoto1 san, thank you for the review. I also confirmed that the parallel build works well. Then, I'll merge this PR.

PR #108, opened 2017-07-20
prepare a core routine for matrix inverse

mn2007:
I think it is no problem to merge this PR even after next release of SALMON. 
But I guess we achive mutual profits by merging this PR. 

This modification is based on PR#92. By the modification, we both use lapack library to calculate matrix inverse in the Broyden's method. 

I recoginze that this modification includes several problems such as position of directory, openmp, etc. But I think this is a first step for us to use shared lapack library. At the moment, matrix size in the Bryoden's method is 8 by 8, so there is no problem to use this routine. 

I've checked dipole mements calculated by GCEED part don't change before and after the modificaion. And I've also checked that a current for single cell calculation in ARTED part doesn't change before and after the modification. 

I don't specify the person who merges because this is the modification for core routines. But I welcome that person who merges and I'm glad if the person is @shunsuke-sato san.


mn2007:
I intended to merge develop branch, and add matrix inverse routine in math. However, large difference of files is displayed. So I decline this PR and prepare another PR.

PR #109, opened 2017-07-21
Modify the density of states calculation routine in 3d periodic case

uemoto1:
## About this PR

This pull request contains some modifications in the density of state (DoS) calculation in ARTED part.
- Provide two different smearing method (Lorentzian and Gaussian).
- Add input parameters to specify "energy range of DoS spectra", "smelling method", "smearing width".
- Use the user specified unit system in `SYSNAME_DoS.out` file
- Add switch to control the zero point to energy
- Change the normalization of DoS from "per Volume" to "per Unit cell"

In this modification, the following input parameters are added in the `analysis` group:
- `out_dos_start` : Lower limit enegy of the DoS spectra
- `out_dos_end` : Upper limit energy of the DoS spectra
- `out_dos_nenergy` : Number of sampling points
- `out_dos_method` : Smearing method `gaussian`/`lorentzian`
- `out_dos_smearing`: Smearing width
- `out_dos_fshift`: Switch ('y'/'n') to shift the Fermi energy to the zero point of electron energy.

## Future work
- Add the corresponding implimentation to the GCEED part.

### DIFF
--- a/ARTED/GS/write_GS_data.f90
+++ b/ARTED/GS/write_GS_data.f90
@@ -16,7 +16,13 @@
 !--------10--------20--------30--------40--------50--------60--------70--------80--------90--------100-------110-------120-------130
 Subroutine write_GS_data
   use Global_Variables
-  use salmon_global, only: out_dos
+  use salmon_global, only: out_dos, &
+                         & out_dos_start, &
+                         & out_dos_end, &
+                         & out_dos_nenergy, &
+                         & out_dos_smearing, &
+                         & out_dos_method, &
+                         & out_dos_fshift
   use salmon_parallel, only: nproc_id_global, nproc_group_tdks
   use salmon_communication, only: comm_is_root,comm_summation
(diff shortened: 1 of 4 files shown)


mn2007:
I checked with 8x8x8 kpoints and the calculated shape of DoS is similar to MatNavi database in NIMS. And the implementation seems to be correct. So I merge this PR.

shunsuke-sato:
@uemoto1 san, thank you for the implementation. But, I would like to ask one thing to you. 
In the present implementation, the range of the energy for DOS must be specified in input. However, the single particle energies are usually unknown before the calculation. Therefore, it is often impossible to give those values from input.

In the previous implementation, the range of the energy is automatically defined by the highest and lowest single particle energies. So, could you re-implement the automatic range finder, which was deleted by this PR?

I would say that the range should be automatically determined by `default`. It is useful for users, and most applications do that.

PR #110, opened 2017-07-24
Fix a bug in calculation of dielectric tensor

uemoto1:
## About This PR
This PR contains a bugfix in dielectric constant (epsilon) in ARTED transverse linear response section.
- Change the background value of the "non-diagonal" component in the dielectric tensor to 0.


mn2007:
It seems to be fine. And I confirmed that currents for a sample of a single cell do not change before and after the modification. So I merge this PR.

uemoto1:
Thank you for merge!

mn2007:
You are welcome. Thanks a lot.

PR #112, opened 2017-07-27
check number of processes compared with the system

mn2007:
I implement function that checks whether total number of proccesses is not too large compared with the system. And I also modify the developer's manual. This modification is related only for the GCEED part. But I checked that it runs successfully for the ARTED and GCEED parts.

Could someone merge this PR?

### DIFF
--- a/GCEED/common/CMakeLists.txt
+++ b/GCEED/common/CMakeLists.txt
@@ -19,6 +19,8 @@ set(SOURCES
     calcVpsl.f90
     check_cep.f90
     check_corrkob.f90
+    check_mg.f90
+    check_ng.f90
     check_numcpu.f90
     conv_core_exc_cor.f90
     conv_p0.f90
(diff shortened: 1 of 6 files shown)


uemoto1:
Thank you @mn2007 san, for implementation. I will review it later.

uemoto1:
Thank you for contribution, and I think it is available to merge to the development branch.

I understood this modification which adds the two check routine  (`check_mg` and `check_ng`) in order to avoid specifying a too large number of processes (or too small grid size). This routine imposes a condition in which the size of real space grid managed by a single processor is much larger than the number required in FDM.

I assume this modification does not have the significant reaction to ordinary calculation and does not touch to the ARTED part.
Since the entire code seems to be successfully build to binary, I will skip the reproduction check in this time.

mn2007:
I appreciate your decision. Thank you, @uemoto1 san.

PR #113, opened 2017-07-27
add a variable "timer_process" in inputoutput.f90

mn2007:
So far, many files for elapsed times are generated in a real-time calculation in the GCEED part. But I think most people were annoying. In addition, a filename like cputime** is confusing because that name is similar to the time displayed by linux command "time". So I add a variable "timer_process" in inputoutput.f90.

By this modification, the master process write elapsed times in standard output in the real-time calculation. If user specify "y" for "timer_process", files named timer_proc** are generated. A default value is "n".

Is it possible to mergy this PR?

### DIFF
--- a/GCEED/rt/real_time_dft.f90
+++ b/GCEED/rt/real_time_dft.f90
@@ -561,73 +561,146 @@ end if
 
 elp3(411)=get_wtime()
 
-write(fileNumber, '(i8)') nproc_id_global
-timeFile = "cputime"//adjustl(fileNumber)
-open(79,file=timeFile)
+if(comm_is_root(nproc_id_global))then
+  write(*,'(a)') "==================== elapsed time ===================="
+  write(*,'(a,f16.8)') "elapsed time bef. reading lda data [s] = ", elp3(402)-elp3(401)
+  write(*,'(a,f16.8)') "elapsed time for reading lda data [s]  = ", elp3(403)-elp3(402)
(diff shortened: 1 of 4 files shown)


mn2007:
I forgot to say one thing. I checked that it runs successfully both for the ARTED part and the GCEED part.

uemoto1:
@mn2007 san, Thank you for development. I think this is a bit complicated and require a bit time for consideration. I will review this PR later...

mn2007:
Thank you for reviewing, @uemoto1 san. I wait.

uemoto1:
Thank you for development, @mn2007.

I understood this modification outputs the elapsed time analysis to the stdout, and add a new (optional) input parameter to enables to export the same information onto a file `timer_proc???`.
I confirmed this modification does not affect to the functionality of ARTED part, and it is available to merge.

As future plans, the file export with a similar format (and input variable) will be implemented in the ARTED section for the compatibility.

mn2007:
I think it's good idea to implement such function in the ARTED part. But I guess it's not urgent task.  By the way, I think other people would not merge this PR. So could you merge, @uemoto1 san? Thanks in avdance.

uemoto1:
I agree that one is not a major problem to solve immediately, this PR is okay to merge.
(Since I was not a suggested reviewer, I have pending to merge this.) 
If it is necessary, I will merge this one now. Thank you for contribution.

PR #114, opened 2017-07-28
modify developer's manual

mn2007:
I modify the developer's manual related to the GCEED part. This modification is applied only to the manual. So I did not check whether it correctly runs.


uemoto1:
Thank you for fixing the developers manual, @mn2007 san.
I think it is available to merge now.

mn2007:

PR #115, opened 2017-07-28
modify developer's manual (2)

mn2007:
I'm not sure that this PR would pass. I modify the developer's manual for ae_shape. I guess the modificated manual represents the current situation correctly.


uemoto1:
I guess the suggested description correctly represents the present implementation, and agree this PR is also acceptable to merge to the developer branch.

As a future plan, the GCEED will also support the pulse functions used in the ARTED part for the compatibility. Since it is not an immediate problem, I think we will resolve this when we have a time enough.

PR #116, opened 2017-07-28
remove variable "rmixrate" in input file

mn2007:
At the moment, a default for mixing density in the GCEED part is Broyden's method and rmixrate isn't used when the Bryoden's method is used. It's confusing to leave this variable, so I remove it. I checked that it runs successfully for the ARTED part and the GCEED part.


uemoto1:
Thank you for modification.
I confirm that this is not significant changes which affect to any other section, and I believe your check results. I guess it is possible to merge the developer branch.

mn2007:

PR #117, opened 2017-07-29
Add salmon_file module

uemoto1:
## About This Pull Request

This modification adds a shared module `salmon_file` which provides an automatic assignation of IO unit: `get_filehandle`. 
And, in this time, the IO unit definition in `inputoutput` module is modified to utilize `get_filehandle` function.

## Usage
- Get the available unit number by `get_filehandle` function just before using `open` method.

```
use salmon_file, only: get_filehandle
implicit none

integer :: fh_example

fh_example = get_filehandle() 
open(fh_example, file='example', status='old')
write(fh_example, *) "hello world"
close(fh_example)
```

## Future Works
- Currently, the assignation of  IO unit is very complicated and distributed to many source files in SALMON. Since I guess this situation is not clear for the maintenance, I recommend replacing the present IO unit assignation to proposed method.

### DIFF
--- a/modules/CMakeLists.txt
+++ b/modules/CMakeLists.txt
@@ -1,4 +1,5 @@
 set(SOURCES
+    salmon_file.f90
     inputoutput.f90
     salmon_global.f90
     salmon_parallel.f90
--- a/modules/inputoutput.f90
+++ b/modules/inputoutput.f90
@@ -23,12 +23,12 @@ module inputoutput
   real(8),parameter :: au_length_aa = 0.52917721067d0
 
 
-  integer, parameter :: fh_variables_log = 801
-  integer, parameter :: fh_namelist = 901
-  integer, parameter :: fh_atomic_spiecies = 902
-  integer, parameter :: fh_atomic_coor = 903
-  integer, parameter :: fh_reentrance = 904
-  integer, parameter :: fh_atomic_red_coor = 905
+  integer :: fh_variables_log
+  integer :: fh_namelist
+!  integer, parameter :: fh_atomic_spiecies = 902
+  integer :: fh_atomic_coor
+  integer :: fh_reentrance
+  integer :: fh_atomic_red_coor
   logical :: if_nml_coor, if_nml_red_coor
 
   integer :: inml_calculation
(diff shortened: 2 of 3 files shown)


mn2007:
Sorry, @uemoto1 san. I can not merge this PR. But other peaple may merge. 
By this implementation, file unit numbers change if new IOs are placed between previous IOs. In the case of fhi98PP, file unit numbers are fixed, and their numbers are written in the manual. fhi98PP is an old program and implementation by @uemoto1 san may be current standard, but I'm not sure it's current standard. Or we can deteremine rules for file unit numbers. In that case, we need to discuss.

uemoto1:
Thank you for reviewing.

I considered your comment, and I found the problem you mentioned would be resolved by introducing rules:

1) Keep insert `get_filehandle` one line before the `open` method:
If single file access closes before opening next file, this automatic assignment works well. However,  in the case of nested file access (as you told),  the collision of assigned number will happen. To resolve this issue, I recommend inserting `get_filehandle` just before the `open`.  (By keeping this rule, the free IO unit will always be assigned. )

2) Keep using n <= 999 number for fixed file IO:
In present implementation, the `get_filehandle` is assigning IO unit number larger than 1000. Therefore, if it is necessary to attach the fixed IO unit for some reason, we can continuously use the number <= 999.

3) Shere the IO unit numbers as global variables (if necessary):
If a few different subroutines share a single file access, I think the rule above is preferred.

And, I have investigated some related first principle codes, I found the automatic assignment is very general technique as below:

- Quantum Espresso (`find_free_unit()`)
https://github.com/QEF/q-e/blob/master/UtilXlib/find_free_unit.f90#L9

- OCTPASS (`io_assign()`)
https://github.com/qsnake/octopus/blob/3f4791704aa10e468f0e47eca0e08d09c1e95dc3/src/basic/io.F90#L245

- ABINIT (`get_free_unit()`)
https://github.com/abinit/abinit/blob/bd4dcfc23387bb6c191e6d13850cbc27f68ef385/src/14_hidewrite/m_io_tools.F90#L125

mn2007:
@uemoto1 san, thank a lot for your investigation. I'd like to use this function. So I merge this PR.

PR #119, opened 2017-07-30
add lorentzian function for DOS and PDOS

mn2007:
Sorry, @uemoto1 san. I need lorentzian function for DOS and PDOS in the GCEED part immediately. So I implement that function to the GCEED part. I checked that calculations for the ARTED part and the GCEED part successfully finished. 

I'd like @uemoto1 san to check this function because you implemented this function to the ARTED part. Could you please check? Since today is holiday, there is no problem if the check is not done today.

### DIFF
--- a/GCEED/scf/CMakeLists.txt
+++ b/GCEED/scf/CMakeLists.txt
@@ -5,6 +5,7 @@ set(SOURCES
     calc_occupation.f90
     calc_pdos.f90
     calc_rho_in.f90
+    check_dos_pdos.f90
     convert_input_scf.f90
     copy_density.f90
     deallocate_sendrecv.f90
--- a/GCEED/scf/calc_dos.f90
+++ b/GCEED/scf/calc_dos.f90
@@ -16,28 +16,52 @@
 subroutine calc_dos
 use salmon_parallel, only: nproc_id_global, nproc_group_grid
 use salmon_communication, only: comm_is_root, comm_summation
-use inputoutput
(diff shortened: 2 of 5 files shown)


uemoto1:
Thank you for implementation.

This PR is very helpful for compatibility. 
I confirmed the choices of `lorentzian` and `gaussian` reproduces the similar results without any significant differences.

For future work, I recommend modifying the header of output files to indicate the unit system correctly and fix the documentation to enables `0d` mode of `out_dos_????` when we have time enough.

However, I think the points above are not major problems, and guess this PR is possible to merge.

mn2007:
Thank you for merging, @uemoto1 san. I may do the modification suggested by you as for future work.

mn2007:
@uemoto1 san, I guess e-mail will be delivered to you by adding comment in this place. So can I confirm what you said? I think unit system is written in header of output dos and pdos files generated in the GCEED part in current situation, and there is no need to modificate. And only I (or we?) have to do is modifying the developer's manual. Is that correct?

uemoto1:
@mn2007 san, I have checked the related part for PR#123, and  I also found the previous recommendation is my misunderstanding. The current implementation, the  file header works well.
I agree there is no necessary to modify the header section.

mn2007:
Thank you for your reply. If PR#124 is merged, I will add same function as PR#124 and modify the manual.

uemoto1:
Thanks!

PR #120, opened 2017-07-30
Enabled to specify grid spacing for ARTED part

shunsuke-sato:
I enabled to use the grid spacing `%rgrid/dl(3)` in input for ARTED. 

SALMON internally computes the number of grid points `%rgrid/num_rgrid(3)` as the nearest integer to the ratio of the size of the cell `%system/al(3)` to the spacing `%rgrid/dl(3)`. 
Therefore, actual grid spacing in calculations is slightly modified.


mn2007:
I checked with num_rgrid=12, 12, 12 and dl=0.85501, 0.85501, 0.85501. Those results are same as previous one. Since deadline is approaching, I merge this PR.

PR #121, opened 2017-07-30
Enabled to use odd numbers of real-space grid points.

shunsuke-sato:
Number of grid points is not necessary to be an even number. Thus, I enabled to use odd numbers of grid points in ARTED part.

### DIFF
--- a/ARTED/common/Hartree.f90
+++ b/ARTED/common/Hartree.f90
@@ -35,7 +35,7 @@ Subroutine Hartree
 !$omp end do
 
 !$omp do private(ix,iy,nz) collapse(3)
-  do nz = -NLz/2,NLz/2-1
+  do nz = -NLz/2,NLz-1-NLz/2
   do iy = 0,NLy-1
   do ix = 0,NLx-1
     f1(ix,iy,nz)=sum(eGzc(nz,:)*rho_3D(ix,iy,:))
@@ -45,8 +45,8 @@ Subroutine Hartree
 !$omp end do
 
 !$omp do private(ix,ny,nz) collapse(3)
-  do nz = -NLz/2,NLz/2-1
-  do ny = -NLy/2,NLy/2-1
+  do nz = -NLz/2,NLz-1-NLz/2
+  do ny = -NLy/2,NLy-1-NLy/2
   do ix = 0,NLx-1
     f2(ix,ny,nz)=sum(eGyc(ny,:)*f1(ix,:,nz))
   end do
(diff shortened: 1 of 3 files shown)


mn2007:
I tried with modifying sc sample to num_rgrid = 13, 13, 13 but calculation stops displaying "Bad grid point". I can not figure out this can be merged. Other people may merge this PR.

uemoto1 (changes requested):
Thank you for developing and reviewing, @shunsuke-sato and @mn2007 

I also detected a same error message. And, I guess this problem is caused by the following part in `initialization.f90`:

```
ARTED/control/initialization.f90:188:if((mod(NLx,2)+mod(NLy,2)+mod(NLz,4)) /= 0)call err_finalize('Bad grid point')
ARTED/control/initialization.f90:201:if((mod(NLx,4)+mod(NLy,4)+mod(NLz,4)) /= 0)call err_finalize('Bad grid point')
ARTED/control/initialization.f90:215:if((mod(NLx,4)+mod(NLy,4)+mod(NLz,4)) /= 0)call err_finalize('Bad grid point')
```

After commenting out the lines above, the calculation of n=13 crystal works and reproduce the similar result of the n=12 case.

shunsuke-sato:
@mn2007 san and @uemoto1 san, thank you for reviewing. I guess both of you used wrong symmetry for the test... 
Because of the grid representation, a crystal symmetry can be broken. So, I guess, in your calculations, SALMON detected the symmetry breakdown and stopped.
So, I would advise you not to use the symmetry function unless it is clear to you.

mn2007:
Thank you for your comment, @shunsuke-sato san. But I guess most users simply change num_rgrid and would not understand what is wrong. So I'm glad if you add recommended variables in stop sentence when users use wrong variables. If you disagree, I check with changing the variable isym and may merge this PR.

uemoto1:
Thank you @shunsuke-sato san, I will re-check it later.

shunsuke-sato:
@mn2007 san, yes, the present error message is not helpful for users. I will modify it in the future.

But, I do not suppose users use this symmetry function because the symmetry function is too tricky, and thus, I do not think that we would provide this function to users in the forthcoming release: the present symmetry function requires specific crystal symmetries, specific atomic coordinates, and a specific laser polarization direction. Furthermore, the number of grid points should be carefully chosen not to break the expected crystal symmetry.
Therefore, I advised you not to use the symmetry function unless it is clear to you.

Also, I would like to emphasize that I provided input files in `ARTED/data/` not for users but for a quick regression test. So, I do not suppose that users use these input files with some modifications.

uemoto1 (approved):

mn2007:
Thank you for your explanation, @shunsuke-sato san. I understand the situation. Therefore, I tried with using default values for "isym" and "crystal_structure". The calculation successfully finishes when num_rgrid = 12, 12, 12, but it stops at the begenning of RT calculation with ommiting error message "core dumped" when num_rgrid = 13, 13, 13. I may give up the review and if @uemoto1 san agrees to merge, there is no problem to be merged.

uemoto1:
Thank you for @shunsuke-sato and @mn2007.

I have confirmed `num_rgrid = 13,13,13` calculation successfully works when the crystal symmetry is fixed to `isym=1` case.
I agree that the implementation of symmetry transforms with odd -number grid will become the complicated modification. For now, the target is limited for the non-symmetric calculation, but it is worth of merging to the develop branch.

When create the documentation of next version, I prefer to add the description; the odd-number grid is allowed only in the case of non-symmetric crystal for future.

mn2007:
Then, I merge this PR. Thank you, @uemoto1 san and @shunsuke-sato san.

Issue #122, opened 2017-07-31
Density of state


shunsuke-sato:
- I moved the following comment from the closed PR since the issue tracker would be better for discussion than the closed PR.

@uemoto1 san, thank you for the implementation for PR #109. But, I would like to ask one thing to you.
In the present implementation, the range of the energy for DOS must be specified in input. However, the single particle energies are usually unknown before the calculation. Therefore, it is often impossible to give those values from input.

In the previous implementation, the range of the energy is automatically defined by the highest and lowest single particle energies. So, could you re-implement the automatic range finder, which was deleted by PR #109 ?

I would say that the range should be automatically determined by `default`. It is useful for users, and most applications do that.

uemoto1:
Thank you for comment, @shunsuke-sato san.

In the previous PR, due to the previous implementation of GCEED-side  DoS calculation ("Gaussian smearing" and "fixed electron energy range"), I assume the similar design is easy to implement for the both ARTED and GCEED side.
In the present implementation has one merit; it is possible to control the adequate plot region and sampling density, when the electronic structure and interested electron energy range are already clear.
On the other hand, as you mentioned,  sometimes it is often difficult to give adequate values as input variables.
An idea I propose is providing an additional variable in inputfile to switch the automatic range specification. (set default as enabling)

If someone can have time enough, please add this routine both for ARTED and GCEED section.

shunsuke-sato:
@uemoto1 san, thank you for your quick response!
I knew the merit for controlling the plot range. However, if the auto-range function does not exist, it would be quite inconvenient for users because SALMON requires the same DFT calculation twice to get DOS. One is to define the energy range, and the other is to compute DOS. Thus, I proposed to employ the auto-range function as default. Of cause, the range can be controlled optionally.

Because you deleted the auto-range function, I first asked you to fix it. But, if you do not do that, I will re-implement it later.

uemoto1:
Okay, I consider a simpler implementation method for the above case.

In this idea, the default value of `out_dos_start` and `out_dos_end` are set to be large negative/positive number which does not appear in the electron eigenenergy.
At the stage of DoS calculation, clip the `out_dos_start`/`out_dos_end` to the lower/upper bound of the calculated eigenenergies.

In this implementation, if the `out_dos_(start|end)` is not specified,  the upper/lower limit of eigen energies are automatically assigned. And in this case, it is not necessary to add one another input variables to switch the automatic detection.

[...]

PR #124, opened 2017-08-01
Add automatic range assumption of DoS output

uemoto1:
## About This Pull Request
This pull request provides the automatic range estimation of DoS output for ARTED and GCEED section. 
- This PR is related to the issue (https://github.com/SALMON-TDDFT/SALMON/issues/122).
- In this implementation, if the input variable `out_dos_start` / `out_dos_end` exceeds the minimal/maximal value of the electron eigenenergies, the upper/lower bound of plot region is clipped inside the electron energy range.
- By default, the plot range is automatically fixed on energy region as above.

### DIFF
--- a/ARTED/GS/write_GS_data.f90
+++ b/ARTED/GS/write_GS_data.f90
@@ -114,21 +114,25 @@ Subroutine write_GS_data
     subroutine dos_write
       use inputoutput, only: t_unit_energy_inv, t_unit_energy
       implicit none
-      real(8) :: vbmax, cbmin, efermi, eshift
+      real(8) :: vbmax, vbmin, cbmax, cbmin, efermi, eshift
       real(8) :: ww, fk, dw
       integer :: iw
-      real(8), allocatable :: dos(:), dos_l(:)  
-      
-      allocate(dos(out_dos_nenergy), dos_l(out_dos_nenergy))
-      
+      real(8) :: dos(out_dos_nenergy), dos_l(out_dos_nenergy)  
+    
+      vbmin = minval(esp_vb_min)
+      vbmax = maxval(esp_vb_max)
+      cbmin = minval(esp_cb_min)
+      cbmax = maxval(esp_cb_max)
+    
       if (out_dos_fshift == 'y') then
-        vbmax = maxval(esp_vb_max)
-        cbmin = minval(esp_cb_min)
         efermi = (vbmax + cbmin) * 0.5d0
         eshift = efermi
       else
         eshift = 0d0
       endif
+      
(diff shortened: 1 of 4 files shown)


shunsuke-sato:
@uemoto1 san, thank you for the implementation! It looks almost fine for me!
I have one comment: The present implementation does not seem to expect the finite electron temperature case, `temperature > 0, which is necessary for (semi) metals. If so, can you fix it?
If it is not trivial for you, please let me know. I will take care of it.
However, I would not have for the moment. So, I'll do it later...

uemoto1:
Thank you for comments,

In the case of finite temperature case, as you mentioned, the determination of the Fermi energy is not a trivial problem. At least, for now, I prefer to add the description about the `out_dos_fshift` switch is only available for the case of zero-temperature calculation.

 I think it is possible to estimate the Fermi level by using the electronic occupation rate. However, I think implementing this estimation both for the ARTED/GCEED part is a bit complicate work. 

If you can handle this problem, I would like request a your help and assign it for you.

shunsuke-sato:
@uemoto1 san, thank you for your quick reply. In the finite temperature calculation, the Fermi energy is internally computed. So, we can just use it. 
Furthermore, in the finite temperature case, the valence and the conduction bands are not well defined. Thus, the range also can be wrong in the present implementation.

OK, I'll work on that. Although I would not have time for the moment, I want to fix it before the release. So, if you have some information about the timing of release, please let me know.

I'll merge This PR later. Thanks a lot!

uemoto1:
I have confirm the problem, and I guess we can use `esp(:,:)` to determine the upper or lower limit of electron eigenenergies. Thank you, @shunsuke-sato san, and looking forward your PR.

shunsuke-sato:
@uemoto1 san, yes, we should use `esp(:,:)`. I'll check and merge the PR tonight.

shunsuke-sato:
@uemoto1 san, I checked the result, but I found the implementation incorrect. The present implementation provides wrong DOS if MPI paralleization is employed.

Only the root process compute `esp_vb_min`, `esp_vb_max`, `esp_cb_min`, and `esp_cb_max`, and the other processes do not have those data. Therefore, `vbmin`, `vbmax`, `cbmin` and `cbmax` in `dos_write` are wrong if MPI parallelization is employed. This is not only a problem for the automatic range assumption, but also for the Fermi energy shift in general.
So, by using `esp` instead of those variables, could you fix the problem, please?

uemoto1:
Thank you for information, @shunsuke-sato san.
I immediately fixed related paralyzation problem for energy estimation as above.

shunsuke-sato:
@uemoto1 san, thank you for fixing the problem. I confirmed that the DOS looks fine. I'll merge this PR.

PR #125, opened 2017-08-01
Fix makefile

shunsuke-sato:
- Removed unused Makefile
- Fixed Makefile for GNU environment


uemoto1 (approved):
Thank you for fixing @shunsuke-sato san.

I have understood this PR removes the previous makefile, and modify to use the native BLAS/LAPACK in the case of GNU compiler. 
I recommend changing the architecture specific header file `/makefiles/arch/gnu` for the convenience to use the generator script. But it is not a significant problem and not necessary to fix in a few days.

I think this PR is available to merge.

Additionary, I consider, in the case of the environment in which Intel and GNU fortran compiler coexist, `mpif90` sometimes call the Intel's one. I prefer to use `mpifc` to execute the gfortran explicitly.

shunsuke-sato:
@uemoto1 san, thank you for your comments! Now I understand these makefiles in /makefiles/ are provided by the script.

I'm not sure whether `mpifc` is so common. For example, in my gnu environment, `mpifc` is not available. Thus, I'm afraid that many users face the same problem, and changed to `mpif90`, which would be rather common.

uemoto1:
Okay, I understood. Presently, I will employ `mpif90` to execute the compiler and continue to investigate the adequate command which used in the other project. Thank you!

PR #129, opened 2017-08-02
modify dos and pdos

mn2007:
Thank you for your implementation, @uemoto1 san. And thank you for your advice, @shunsuke-sato san. But I found that top and bottom dos are cut when I use lorentzian function for C2H2 case. This is what @yabana sensei said. I'd like to modify the code as following. 

1. I enlarge energy ranges for the DOS and PDOS plot. 
2. Since out_dos_start and out_dos_end in std input is no longer used, I remove them. But they remain as SALMON global variables.
3. According to the coding rules, I modify from out_dos_nenergy to iout_dos_nenergy. 

People who check may disagree above modification. In that case, I'll modify. So please do not hesitate to comment on this PR.

### DIFF
--- a/ARTED/GS/write_GS_data.f90
+++ b/ARTED/GS/write_GS_data.f90
@@ -19,7 +19,7 @@ Subroutine write_GS_data
   use salmon_global, only: out_dos, &
                          & out_dos_start, &
                          & out_dos_end, &
-                         & out_dos_nenergy, &
+                         & iout_dos_nenergy, &
                          & out_dos_smearing, &
                          & out_dos_method, &
                          & out_dos_fshift
@@ -117,7 +117,7 @@ Subroutine write_GS_data
       real(8) :: vbmax, cbmin, emax, emin, efermi, eshift
       real(8) :: ww, fk, dw
       integer :: iw
-      real(8) :: dos(out_dos_nenergy), dos_l(out_dos_nenergy)  
+      real(8) :: dos(iout_dos_nenergy), dos_l(iout_dos_nenergy)  
     
       if (comm_is_root(nproc_id_global)) then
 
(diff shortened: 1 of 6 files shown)


mn2007:
I forgot to say one thing. I tested for the ARTED part and the GCEED part as much as I consider.

uemoto1:
Thank you for contribution, @mn2007 san.

I understand your point and agree to use `iout_dos_nenergy` to obey of the coding rules.   This PR also modified estimation method which provides the much wider range than all electron eigenenergies. I agree it is better than the previous one to understand the entire DoS structure from one plotted graph.
On the other hand, I guess there is a merit to specify the upper and lower bound for usabilities. 

Therefore, I have an idea to add the lines below:
```
if (out_dos_start <= 1.d+10) out_dos_start = ene_min-0.25d0*(ene_max-ene_min)
if (out_dos_end >= 1.d+10) out_dos_end = ene_max+0.25d0*(ene_max-ene_min)
```
or
```
out_dos_start = max(out_dos_start, ene_min-0.25d0*(ene_max-ene_min))
out_dos_end = min(out_dos_end, ene_max+0.25d0*(ene_max-ene_min))
```
.
 
I prefer using the implementation above. it makes possible to coexist both of  @mn2007 san's method and previous range specification in the same code.

I think this PR is mainly available to merge. 
If you need immediate merging, please told it to me.

mn2007:
Thank you for your comment, @uemoto1 san. I would like to implement latter one you suggested. So could you wait until I implement?

uemoto1:
Thank you for consideration, I will looking forward the next upload.

mn2007:
I finished the implementation. I checked with setting and not setting out_dos_start and out_dos_end for both the ARTED part and the GCEED part, and I obtained results which are satisfied.

mn2007:
If @uemoto1 san judges that it's acceptable to merge, could you merge this PR immediately? Thanks in advance.

uemoto1:
Thank you @mn2007 for modification.
I have checked the DoS calculation in ARTED part works well, and it is available to merge to develop branch.

PR #130, opened 2017-08-02
change names of std output files

mn2007:
On wiki, std output files "???.out" are written in the periodic system calculation. So I'd like to modify names of std output files for the isolated system calcualtion to match ones for the periodic system calculation.  I didn't any check because this is a mere modification for shell scripts.


uemoto1:
Thank you @mn2007 san, I will merge this pull request.

mn2007:

PR #131, opened 2017-08-03
modify info file

mn2007:
I modify a .info file generated in the GCEED part. Just to make sure, I checked that the calculations for the ARTED part and the GCEED part successfully finished.


uemoto1:
Thank you for developing, @mn2007. I will check it later.

uemoto1:
I confirmed this PR modifies the data format of `SYSNAME.info` file. And, I have checked the output file is quantitive equivalent to the previous results. I think this PR is available to merge.
Thank you for contribution, @mn2007 san.

PR #132, opened 2017-08-03
Fix a header of makefile template

uemoto1:
This PR modify the header section of gnu makefile (`/makefiles/arch/gnu`) to reproduce current `Makefile.gnu` by generator script.
Since this modification has no affection to build process and binary file, someone can merge this PR without any testing.


mn2007:
There seems to be no problem. So I merge this PR.

PR #133, opened 2017-08-04
change unit of output files for avs

mn2007:
I change unit of output files for avs. The unit is changed from a.u. to A/eV/fs. The reason why I modify is that A/eV/fs is familiar to us. I also add header to identify unit. In future, I may modify to correspond both unit. For cube files, length is generally specified by a.u. So I don't change cube files.  

I checked the calculation successfully finished as always. And I checked that time propagation of difference of density is correcly displayed. 

It may be difficult to check this PR. But I'd like to pass this PR before next release of SALMON. Could you please check this PR?

### DIFF
--- a/GCEED/common/writeavs.f90
+++ b/GCEED/common/writeavs.f90
@@ -13,12 +13,13 @@
 !  See the License for the specific language governing permissions and
 !  limitations under the License.
 !
-subroutine writeavs(fp,suffix,tmatbox_l)
+subroutine writeavs(fp,suffix,header_unit,tmatbox_l)
   use salmon_parallel, only: nproc_id_global
   use salmon_communication, only: comm_is_root
   use scf_data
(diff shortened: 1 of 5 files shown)


uemoto1:
Thank you for developing, @mn2007. Since we do not have the AVS express package, it is so hard to figure out the result is correct or not. Therefore, I will check the PR within the conventional calculation.

uemoto1:
I have checked the conventional calculation with `analysis.out_dns='y'` and `analysys.format3d='avs`. I think the conventional calculation are working, and the `dns.out` file is successfully exported. We cannot be not sure for the visualization, however I'd like to believe your report. I think it is available to merge to developer branch. Thank you for developement.

mn2007:
I appreciate your detailed checks. Thanks a lot, @uemoto1 san.

Issue #103, opened 2017-07-12
Comprehensive implementation of Bessel function


shunsuke-sato:
The present implementation of the Bessel function in `modules/salmon_math.f90` is valid only for relatively small `x`. To avoid wrong use of this function, we need a comprehensive implementation as soon as possible.

PR #134, opened 2017-08-04
Modify the bessel function j1(x)

uemoto1:
## About This PR

This pull request provided the implementation of the first order Bessel function and related to the https://github.com/SALMON-TDDFT/SALMON/issues/103.

For the convenience, the comparison between the modified function and reference [scipy.special.j1](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.j1.html) is attached [here](https://www.dropbox.com/s/8mhyp2pg0tnciuv/uemoto0804_salmon_math_j1.png?dl=0).

Currently, this function j1(x) in the large-x region is not used in this program. Therefore, this PR has less priority to merge. If someone has the time enough, please check it.

### DIFF
--- a/modules/salmon_math.f90
+++ b/modules/salmon_math.f90
@@ -165,22 +165,73 @@ contains
   
 !--------------------------------------------------------------------------------
   !! 1st order Bessel function
+  ! The algorithm and numerical table are implemented based on the reference:
+  ! Methods and Programs for Mathematical Functions, Prentice-Hall, 1989
   real(8) function bessel_j1_salmon(x) result(y)
     implicit none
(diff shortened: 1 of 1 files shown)


mn2007:
I'll add comments. I merge this PR.

mn2007:
I compared the implemented function with following three routines that calculate j1(x)
1. http://keisan.casio.jp/exec/system/1173919505
2. SSL2 developed by fujitsu
3. http://www.kurims.kyoto-u.ac.jp/~ooura/bessel.html

and they almost matched. So I merged this PR.

PR #136, opened 2017-08-05
Add the common GS output files

uemoto1:
## About This Pull Request

This modification provides the subroutines of the common output file format for SALMON-TDDFT. This PR experimentally implements the output for GS calculation. 
I employ the file formats discussed on **SALMON June meeting**:
https://www.dropbox.com/s/43qyhxj0n5y7a5z/uemoto0805_output_format.pdf?dl=0
(Attached is document distributed in previous meeting)

The ordinary output files (e.g. `SYSName_band.out` ...) have also remained for now to keep the compatibility of the old calculation results.

##  Output Files
- `SYSName_k.data` : describe the k-point coordinates (only in 3d mode)

| Column | Detail | Unit |
|----------|------------|------|
| 1 | k-point index | integer starts from 1 |
| 2:4 | kx, ky, kz coordinates | 1 / length |
| 5 | k-point weights | dimensionless |

- `SYSName_eigen.data` : describe the eigenenergies and occupation number

| Column | Detail | Unit |
|----------|------------|------|
| 1 | k-point index | integer starts from 1  |
| 2 | band index | integer starts from 1  |
| 3 | eigenenergy  | energy |
| 4 | occupation | dimensionless |

## Additional Work
- In this time, I add the function (`open_filehandle`) to `salmon_file` module, which provides the similar functionality of `fopen` (in C) or `open` (in Python).
```
fh=open_filehandle("filename.txt", status="old")
```
- `open_filehandle` opens a file with assigning the free io unit number automatically, and it is equivalent to `get_filehandle` and `open` statement.
```
fh = get_filehandle()
open(fh, "filename.txt", status="old")
```

## Future Work
- Add the similar output subroutines for GCEED section
- Add the common output formats of RT calculation

### DIFF
--- a/ARTED/GS/write_GS_data.f90
+++ b/ARTED/GS/write_GS_data.f90
@@ -108,9 +108,11 @@ Subroutine write_GS_data
 
   
   if(out_dos == 'y')call dos_write
-  call write_k_data
-  call write_eigen_data
 
+  if (comm_is_root(nproc_id_global)) then
+    call write_k_data
+    call write_eigen_data
+  end if
   return
 
   contains
--- a/ARTED/GS/write_GS_data.f90
+++ b/ARTED/GS/write_GS_data.f90
@@ -23,8 +23,11 @@ Subroutine write_GS_data
                          & out_dos_smearing, &
                          & out_dos_method, &
                          & out_dos_fshift
+  use inputoutput, only: unit_length, ulength_from_au, &
+                        & t_unit_energy_inv, t_unit_energy
   use salmon_parallel, only: nproc_group_global, nproc_id_global, nproc_group_tdks
   use salmon_communication, only: comm_is_root,comm_summation, comm_bcast
+  use salmon_file, only: open_filehandle
   implicit none
   integer ik,ib,ia,iter,j
 
@@ -105,6 +108,8 @@ Subroutine write_GS_data
 
   
   if(out_dos == 'y')call dos_write
+  call write_k_data
+  call write_eigen_data
 
   return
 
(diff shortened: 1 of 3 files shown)


mn2007:
I'd like to check for this PR. Maybe I'll check next Monday. Sorry for inconvenience.

uemoto1:
Thank you! I'm looking forward it.

mn2007:
Thank you for your development, @uemoto1 san. But from I've seen, it seems that several problems exist in this modification. These might be my misunderstandings, so please do not hesitate to disagree my opinion. 

I checked both the ARTED part and the GCEED part. For the GCEED part, calculations successfully finished. And for the ARTED part, generated current is same as previous one. 

I run the ARTED part in the intel environment. The problems what I find are following three.
1. Si_k .data is not generated.
2. Somehow k-point index generated in Si_eigen.data is 1 to 12 although number of k points is 64. 
3. It seems that there is no description "if (comm_is_root(nproc_id_global)) then" for the part in this modification, so access from many processess might happen at the same time if users do a calculation for a large system.  

Thanks in advance for your consideration.

mn2007:
I forgot to say one thing. Since today is holiday, my reply will be tomorrow at the earliest.

mn2007:
Sorry, I forgot to say another thing. The problem no. 1 may caused by my environment. If @uemoto1 san can generate Si_k.data, you can ignore this problem.

mn2007:
I understand the problem no. 2. In the ARTED, symmetry is considered. So number of k points in this calculation is 12. Sorry for my misunderstanding.

uemoto1:
Thank you for reviewing, it is very practical mistake. I send a bug fix.

mn2007:
I confirmed that problems have been solved with running the code. So I merge this PR. Thanks a lot, @uemoto1 san.

PR #143, opened 2017-08-09
Reduced GNU Makefile size.

yhirokawa-ccs:
I renamed makefiles/template.txt to makefiles/make.body. All makefiles import it.


uemoto1:
Thank you @yhirokawa san.
Using `include` in make is very helpful to simplify the makefile.
I prepare the makefile to execute `make` command with moving the `makefiles/Makefile.arch` to the top of source tree. Therefore, it is necessary to change the path to specify `make.body` file.

`include $(dir $(abspath $(lastword $(MAKEFILE_LIST))))/makefiles/make.body`

Can you modify the line of include path as above?

yhirokawa-ccs:
@uemoto1 san, `$(dir $(abspath $(lastword $(MAKEFILE_LIST))))` means that absolute path of this makefile, so this makefile can be executed on the top of source tree.

    $ pwd
    /home/yhirokawa/SALMON
    $ make -f makefiles/Makefile.intel
    ...

uemoto1:
Thank you, @yhirokawa san. I understand your point.

PR #148, opened 2017-08-09
Modified matrix inversion.

shunsuke-sato:
I modified the matrix inversion routine.


mn2007:
Thank you for your development, @shunsuke-sato san. I'll check later. If there is no problem, I merge it.

PR #151, opened 2017-08-10
solve the dependency in cmake

mn2007:
I solve the depencency in cmake, which appears in src/ARTED/FDTD. Please merge this PR.


uemoto1:
Thank you @mn2007 san, I think this modification is correct and worth to merge the developer branch.

PR #156, opened 2017-08-14
Suppressed compiler warnings.

yhirokawa-ccs:


mn2007:
Thank you, @yhirokawa san. But I got in trouble with the build with cmake on FX10. Following errors were displayed at the final.
<pre>
[100%] Linking Fortran executable ../salmon.cpu
ARTED/RT/libarted_rtl_RT.a(hamiltonian.f90.o):(.data+0x30): undefined reference to `salmon_unusedvar_'
GCEED/modules/libgceed_rtl_modules.a(hpsi2.f90.o):(.data+0x48): undefined reference to `salmon_unusedvar_'
xc/libxc_rtl.a(exc_cor.f90.o):(.data+0x0): undefined reference to `salmon_unusedvar_'
make[2]: *** [salmon.cpu] Error 1
make[1]: *** [src/CMakeFiles/salmon.cpu.dir/all] Error 2
make: *** [all] Error 2
</pre>
It seems to be difficult for me to solve this problem. Could you fix it? Or, I'll try to fix it. In that case, it will take a long time ... .

mn2007:
Could ARTED users comment on this PR if there is any problem for you? If not, I'll check the ARTED part and the GCEED part and merge this PR.

uemoto1:
Thank you for developing and reviewing

I found there are no problems with generated binaries.
The claimed error message seems to originate from an implementation particular to Fujitsu compiler, which scans the existence of the subroutines even if it doesn't use in the program.

I guess the solution is adding the definition of `salmon_unsedvar`  subroutine into the code. (Since I don't have SC's account, I can't verify this problem.)

The built binary works well, and it is available to merge.

uemoto1 (changes requested):
Thank you for developing and reviewing

I found there are no problems with generated binaries.
The claimed error message seems to originate from an implementation particular to Fujitsu compiler, which scans the existence of the subroutines even if it does not used by the preprocessors.

I guess the solution is adding the simple definition of `salmon_unsedvar`  subroutine to somewhere in the code. (Since I don't have Fujitsu platform's account, I can't verify this problem.)

The built binary works well, and it is available to merge (excluding the problem in Fujitsu)

mn2007:
I understand the situation and agree with @uemoto1 san. @yhirokawa san, could you please add `salmon_unsedvar` subroutine somewhere? Otherwise, I cannot run SALMON on fujitsu environment such as FX10 or K computer if this PR is merged.

yhirokawa-ccs:
Thank you for reviewing. I fixed it.

In compile time, most of compilers eliminate `if(.false.) call salmon_unusedvar(VAR)` because it is unreachable codes.
Fujitsu compiler not eliminate the unreachable codes, but it is not wrong behavior.

uemoto1 (approved):

uemoto1:
Thank you @yhirokawa san, for modification.
Although it is not sure for Fujitsu platform, I think it works in my environment.

mn2007:
I succeeded to build and run, and the calculation for the GCEED part successfully finished on FX10. Thank you so much, @yhirokawa san and @uemoto1 san. I merge this PR.

PR #158, opened 2017-08-20
Refactors CMake script tree.

yhirokawa-ccs:
I added `cmakefiles` directory for CMake scripts as with GNU Make scripts.
`platform` directory renamed `cmakefiles/Platforms`, but build process is same manner as before.


uemoto1:
Thank you @yhirokawa san, I will check it later.

mn2007:
I checked for the GCEED part and build and run for the part successfully finished on FX10. @uemoto1 san, could you check the ARTED part? And if there is no problem, could you merge this PR?

uemoto1:
Thank you for contiribution. I think this PR is available to merge.

PR #160, opened 2017-08-21
Fix minor problems in document

uemoto1:
## About This Pull Request

This PR is the minor change of the document `input_variables.md` which contains:
- Fix the option of `propagater`
- Fix the description of `file_kw`

Since this PR doesn't modify any source file, it is not necessary to varify the build and calculation.

### DIFF
--- a/manual/input_variables.md
+++ b/manual/input_variables.md
@@ -422,7 +422,7 @@ Choice of Propagator.
 <code>middlepoint</code> is an propagator
 with the Hamiltoinan at midpoint of two-times.
 
-<code>etrs</code> is Enforced time-reversal symmetry propagator.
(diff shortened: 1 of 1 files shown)


mn2007:
For num_kgrid, the description is consistent with the code. And for propagator, modified description is correct. So I think this PR is able to merge. But I think it's better to use small capital at the first letter of "Enforced time-reversal symmetry propagator". @uemoto1 san, will you modify or not? I'll merge whichever you choose.

uemoto1:
Thank you @mn2007 san, I fixed the issued point as above.

mn2007:
Thank you for your reply and the modification, @uemoto1 san. I merge this PR.

PR #161, opened 2017-08-22
Fix minor problem in document

uemoto1:
## About This Pull Request
This pull request contains the minor fix of the `input_variables.md` mistakes, which found in the developement of inputfile converter:
- Fix the description of `control`/`directory` variable
This modification does not touch the source codes, therefore it is not necessary to check the build and calculation. 

## Future Work
- Update the documentation (for v.1.0.0) in website as the current version.

### DIFF
--- a/manual/input_variables.md
+++ b/manual/input_variables.md
@@ -85,7 +85,7 @@ Default is <code>-1</code> sec.
 Default is <code>default</code>.
 </dd>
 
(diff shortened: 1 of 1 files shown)


mn2007:
This modification is consistent with salmon_global.f90, so I merge this PR. Thank you, @uemoto1 san.

PR #162, opened 2017-08-23
Minor change in CMake files.

yhirokawa-ccs:


uemoto1:
Thank you for developing, @yhirokawa san.
I think this CMake files works well in my environment and it is possible to merge to developer branch.
I will also checks the next PR#163, and merge it later.

PR #163, opened 2017-08-23
Fix configurations of optimization option in CMake.

yhirokawa-ccs:
I refactored the configurations of optimization option in CMake.
+ renamed options
+ `configure.py`
    + reorganized option groups
    + added enable/disable switch


uemoto1:
I confirmed the modified program can be successfully compiled to the binary in my environment.
And, I grasped this pull request changes the name of CMake variables (e.g. `ENABLE_ARRAY_PADDING` to `ARRAY_PADDING`),  and changed some command line switches in `configure.py` (e.g. `--array-padding` to `--(enable|disable)-array-padding`).
I will reflect them to the installation manual, later.

Thank you for contribution.

PR #164, opened 2017-08-23
modify mistake about Ecos2

mn2007:
There was a mistake that "Esin2sin" still remained in the GCEED part. So I modify it. I confirmed that results don't change as usual.


uemoto1:
Thank you for developing @mn2007 san.
I will check it later.

uemoto1:
I think this modification is available to merge to the development branch.

Thank you for the contribution!

mn2007:
Thank a lot for your quick check, @uemoto1 san.

PR #165, opened 2017-08-25
Unified hpsi subroutine (test version)

syamada0:
I add `--hpsi_test` (default=off) option for testing the unified hpsi subroutine.

new subroutines:
src/common/hpsi.f90
src/common/stencil.f90
src/common/update_overlap.f90

wrappers:
src/ARTED/common/hpsi.f90     -->  src/ARTED/common/hpsi_test.f90
src/GCEED/modules/hpsi2.f90   -->  src/GCEED/modules/hpsi2_test.f90
src/GCEED/rt/taylor.f90               -->  src/GCEED/rt/taylor_test.f90


mn2007:
Thank you for your contribution, @syamada0 san. I'll check the GCEED part later. Maybe I can reply some comments next Monday.

uemoto1:
add to whitelist

uemoto1:
The comment above is an execution command to Jenkins

shunsuke-sato:
@syamada0 san, thank you for your implementation. I'll review it this weekend.

shunsuke-sato:
@syamada0 san, thanks again for your implementation. I had a quick check and confirmed that it reproduces the original result.
I have a short question. When I tested the code, I found that the Hamiltonian operation gets 30 - 40 % slower than the original code for ARTED part. Do you expect it?
If yes, please let me know. Then, I will continue the review and approve the PR. Probably, re-optimization would be a future task.

syamada0:
@shunsuke-sato san and @mn2007 san, thank you for reviewing.

@shunsuke-sato san,
Yes, such slowdown is expected because these subroutines are not optimized yet.

shunsuke-sato:
@syamada0 san, thank you for your quick response! I agree the optimization is not needed in this step.Then, I reviewed and approve it!
Thanks for your implementation.

shunsuke-sato (approved):

mn2007 (changes requested):
Could you remove declarations of ispin and replace Mps to variables such as Mps_hpsi in both hpsi2_test.f90 and taylor_test.f90?

mn2007:
I'm not familiar with the review so comment above is given at first. I'll give detailed explanation. So could you wait until I explain? Sorry for inconvenience.

mn2007:
Thank you for your large contribition, @syamada0 san. I give comments. 

I usually use a FX10 machine in IMS (Institute for Molecular Science) to check results calculated by SALMON. Sometimes Fujitsu compilers don't pass the build although intel compilers do pass it. In the present case, variables `ispin` and `Mps` is already declared in other place. So the build doesn't succeed on the FX10. The easiest solution is what I gave above, but you may avoid this problem by other ways. So could you fix this problem?

If the problem is solved, the results calculated in the GCEED part are reproduced. The calculation cost with `--hpsi_test` increases 5 %. But I don't care it at the present.

syamada0:
@shunsuke-sato san, thank you for your quick check and approval.

@mn2007 san, thank you for pointing it out, and I fixed the problem. 
Could you check the correction?

mn2007 (approved):
I approve this PR.

mn2007:
Thank you for your modification, @syamada0 san. Modified code reproduces previous results. @shunsuke-sato san and I have approved this PR, So I merge this PR.

PR #166, opened 2017-08-28
fix md option and implement geometry opt in arted side

ayamada224:
(1) Bug fix for md option (use_ehrenfest_md) --- allocation & update of zJxyz and zKxyz etc
(2) Implement geometry optimization -- add optimization subroutine, some changes in ground state parts (set convergence threshold, print level, initial guess etc only for optimization). Also added trajectory output files for md and opt. 
(Both are in arted side)

### DIFF
--- a/src/ARTED/control/control_sc.f90
+++ b/src/ARTED/control/control_sc.f90
@@ -480,6 +480,7 @@ subroutine calc_opt_ground_state
      if(flag_geo_opt_atom(i)/='y')then
         if(comm_is_root(nproc_id_global)) &
         &  write(*,*)'ERROR: flag of geometry opt of all atoms must be y'
+        call end_parallel
         stop
      endif
   enddo
@@ -529,6 +530,7 @@ subroutine calc_opt_ground_state
   if(fave .le. fave_conv) then
     if(comm_is_root(nproc_id_global)) &
     &  write(*,*) " Mean force is enough small: stop calculation"
+    call end_parallel
     stop
   endif
 
--- a/src/ARTED/control/control_sc.f90
+++ b/src/ARTED/control/control_sc.f90
@@ -867,12 +867,16 @@ end subroutine calc_opt_ground_state
 subroutine write_xyz(NI,Rion,comment,action)
   use inputoutput, only: au_length_aa
   use salmon_global, only: iflag_atom_coor,ntype_atom_coor_cartesian,ntype_atom_coor_reduced
+  use salmon_parallel, only: nproc_id_global
+  use salmon_communication, only: comm_is_root
   implicit none
   integer :: ia,NI,unit_xyz=200,unit_atomic_coor_tmp=201
   real(8) :: Rion(3,NI)
   character(3) :: action
   character(100) :: char_atom,atom_name
   character(*) :: comment
+
+  if(.not. comm_is_root(nproc_id_global)) return
 
   select case(iflag_atom_coor)
   case(ntype_atom_coor_cartesian)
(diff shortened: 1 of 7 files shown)


uemoto1:
add to whitelist

uemoto1:
The above comment is the forcing execution command to Jenkins

shunsuke-sato:
@ayamada224 san, thank you for your implementation. 
- I confirmed that this patch passes the regression tests in the `example` directory. I also tested the geometry optimization for Si from a displaced initial atomic positions, and the code produced a reasonable result.

- Although I had a quick review, I could not carefully review this patch, and it would be a bit difficult to manage the time this week. So, I approve this patch, but would like to ask @uemoto1 san to review this patch closely.

- Finally, I found a minor bug in 'write_xyz' routine, where all the MPI processes write down the data in the same file, and the data were broken. I have fixed the routine and sent a pull-request to your branch. So, please merge it.

Thanks for your contribution.

shunsuke-sato (approved):

uemoto1:
Thank you for developing and reviewing, @ayamada224 san and @shunsuke-sato, and sorry for late reply. I will review this modification later.

src/ARTED/control/ground_state.f90
```diff
@@ -135,10 +141,22 @@ subroutine calc_ground_state
           end if
           write(*,*) '-----------------------------------------------'
        end if
+
+       !(exit if energy difference is below threshold: only if set convrg_scf_Eall>0)
+       if( abs(Eall_GS(iter)-Eall_GS(iter-1)) .le. convrg_scf_ene ) then
+          if(kbTev >= 0d0) write(*,*) "sorry, occ.out was not generated" !(fix if need)
```
uemoto1: The condition of `comm_is_root(nproc_id_global)` will be required here.

src/ARTED/control/control_sc.f90
```diff
@@ -426,4 +443,499 @@ subroutine reset_rt_timer
...
+  call cal_mean_max_forces(NI,force,fave,fmax)
+  if(comm_is_root(nproc_id_global)) then
+     write(*,135) 0, 0, Eall, 0d0
+     write(*,120) " Max-force=", fmax, "  Mean-force=", fave
+  endif
+  !(initial check of convergence: if force is enough small, exit before opt)
+  if(fave .le. fave_conv) then
+    if(comm_is_root(nproc_id_global)) &
+    &  write(*,*) " Mean force is enough small: stop calculation"
+    stop
```
uemoto1: I prefer to call `end_parallel` subroutine before stop the program.

src/ARTED/control/ground_state.f90
uemoto1: The condition of `comm_is_root(nproc_id_global)` might be required here.

uemoto1 (approved):
Thank you for developing @ayamada224 san.

I grasped this modification contains the bug fix of `use_ehrenfest_md='y'`, which fix the problem of deallocation of `zJxyz`/`mJxyz` in `not_initial` case. This PR also append the new calculation mode of `use_ehrenfest_md='y'` options for the ARETD part, it adds subroutine of `calc_opt_ground_state`  a few modifications to the `calc_ground_state` in order to call from the `calc_opt_ground_state` in each searching iteration.

Accordings to @ayamada224  san's recent calculation report, this optimization method successfully produces the atomic positions of some molecules. I agree this implementation works well.

I think it is available to merge the developer branch.

uemoto1:
Thank you for the implementation, @ayamada224 san.
I think this branch is available to merge, and I will merge it.

PR #169, opened 2017-09-01
Add persistent communication routines.

yhirokawa-ccs:


mn2007:
I examined all examples including a multi-scale calculation and confirmed that results don't change. So I merge this PR.

PR #170, opened 2017-09-01
ARTED part: backup_routines.f90 and timer.f90 moves to misc dir.

yhirokawa-ccs:
These are stand-alone modules.


mn2007:
I examined all examples without a reentrance option used in the ARTED part. And I confirmed that results don't change with this modification. In addition, a log by a time routine seemed to be successfully generated in my view. However I'm not sure how to use backup function in the ARTED part. 

It would be helpful for me that someone teachs me how to use the function or checks this PR. But if there are no comments on this PR by afternoon on Monday, I'll start to try again to use backup function by myself.

shunsuke-sato:
@yhirokawa san, thank you for the implementation, and @mn2007 san, thank you for reviewing it.
I think I should check the backup part. So, I'll check and review it tonight, and approve the request.

shunsuke-sato:
I just realized that this is just the file transfer. So, nothing affects results once properly compiled. I approve and merge this request.

mn2007:
Thank you for the decision, @shunsuke-sato san, and thank you for the development @yhirokawa san.

PR #172, opened 2017-09-04
Modified the definition of the time delay.

shunsuke-sato:
I modified the definition of the pump-probe time-delay in ARTED part.

While the conventional definition is the delay between the peaks of the envelopes, SAMON defines it as the delay of the pulse heads. It would be better to use the conventional definition, and thus, I fixed it.

Once this patch is merged, I would like to ask @mn2007 san to fix the GCEED part to fit the conventional definition.

### DIFF
--- a/src/ARTED/RT/init_Ac.f90
+++ b/src/ARTED/RT/init_Ac.f90
@@ -171,7 +171,7 @@ Subroutine init_Ac
     end select
 
     do iter=0,Nt+1
-      tt=iter*dt - 0.5d0*pulse_tw2 - T1_T2
+      tt=iter*dt - 0.5d0*pulse_tw1 - T1_T2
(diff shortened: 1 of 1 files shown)


mn2007:
I understand the situation, @shunsuke-sato san. I don't have a plan to use the pump-probe function at the moment, so I'll follow your definition after this PR is merged. I think it is better to merge this PR by other people. But if no one checks this PR, I'll check and merge.

uemoto1:
Thank you @shunsuke-sato and @mn2007 san for developing and reviewing.
I will also review it for the effect to ARTED part later.

uemoto1:
Thank you for implementation, @shunsuke-sato san, I will confirm this pull request.

shunsuke-sato:
@mn2007 san and @uemoto1 san, thank you for your quick review.

PR #173, opened 2017-09-05
modify the definition of the time delay

mn2007:
I modify the definition of the time delay in the GCEED part based on PR#172. I'd like to ask @shunsuke-sato san to merge this PR if this modification is correct.

### DIFF
--- a/src/GCEED/rt/calcVbox.f90
+++ b/src/GCEED/rt/calcVbox.f90
@@ -136,7 +136,7 @@ SUBROUTINE calcVbox
         end do
         end do
       end if
-      if(dt*dble(itt)-t1_t2 >= 1.d-12 .and. dt*dble(itt)-t1_t2 <= pulse_tw2)then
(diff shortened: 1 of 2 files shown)


shunsuke-sato:
@mn2007 san, thank you for implementation. But, it seems to be incorrect.
The pulse duration is defined not in `calc_env_trigon` but in `calcVbox`. Thus, the condition in `calcVbox` should be properly modified to shift the pulse in time.

mn2007:
@shunsuke-sato san, thank a lot for your review. I'll consider your point, and maybe reply on today's evening.

mn2007:
I also modify the pulse duration in calcVbox.f90. @shunsuke-sato san, could you check?

shunsuke-sato:
@mn2007 san, thank you for fixing the code. I reviewed it and it looks fine. So, I merge it.

mn2007:
@shunsuke-sato san, thank you for merging.

PR #174, opened 2017-09-06
Fix GNU Makefile

uemoto1:
## About This Pull Request
This pull request contains the update of GNU makefiles, which is not working presently. In this time,  the Makefile are fixed in order to follow up the modification of the directory location proposed on PR#165.

I have already tested this new makefile only for `intel` and `gcc` environment in Oakforest-PACS. However, I have not checked how it works in Fujitsu environment. Please, someone check this?

### How to build by GNU Makefile
Enter to the makefile directory:
```
$ cd SALMON/makefiles
```
and execute `make` command with the platform specified makefile:
```
$ make -f Makefile.PLATFORM
```

### Supported Platform
- fujitsu
- gnu
- gnu-without-mpi
- intel
- intel-avx
- intel-avx2
- intel-knc
- intel-knl
- intel-without-mpi

### DIFF
--- a/makefiles/Makefile.fujitsu
+++ b/makefiles/Makefile.fujitsu
@@ -11,4 +11,6 @@ LIBSCALAPACK = -SCALAPACK -SSL2BLAMP
 MODULE_SWITCH = -M
 COMM_SET =
 
-include $(dir $(abspath $(lastword $(MAKEFILE_LIST))))/make.body
+SALMON = $(abspath $(dir $(lastword $(MAKEFILE_LIST)))/../)
+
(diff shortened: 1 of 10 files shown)


mn2007:
Thank you for your implement, @uemoto1 san. I'll check this in the Fujitsu environment.

mn2007:
Since the build succeeded and the code reproduced previous results on the Fujitsu environment, I merge this PR.

uemoto1:
Thank you @mn2007 san, for the prompt reply and quick merge!

PR #175, opened 2017-09-06
solve a problem of openMP with double nest

mn2007:
There is a problem of openMP with double nest which appears in the ARTED side. This problem affects only for Fujitsu environment. Most people may not use the Fujitsu environment, but I often use it. So I fix the problem. If there are better solutions compared with this one, I decline this PR.


uemoto1:
Thank you @mn2007 san. I will check this section.
Since the modification is a bit complicate, I need have a moment to answer the decision...

uemoto1:
Thank you for bugfix and investigation, @mn2007 san, @shunsuke-sato san, and @yhirokawa san.
I understand the point of the problem, and I agree it is necessary to merge this PR to develop branch.

PR #176, opened 2017-09-06
Optimizing halo communications in GCEED part.

yhirokawa-ccs:
I optimized halo communications (peer-to-peer communications) in GCEED part as follows:

1. optimize pack/unpack routines.
2. use persistent communication of MPI.

I checked the performance results under Xeon Phi 16 nodes, this PR reduced 20% of total execution time when RT calculation.

## Verification of results

I verified calculation results with C2H2 molecule example under Xeon Phi as following process distributions:

1. Xeon Phi executes with 16 MPI processes
2. `nproc_ob` sets `2`.
3. `nproc_domain` sets `2,2,2`
4. `nproc_domain_s` sets `4,2,2`

## Note

I will send PR for reducing and unifying halo communications after merged it.

## Suggestion
`C_sendrecvh` and `sendrecv_groupob_ngp` are never used, so I suggest that remove it.

### DIFF
--- a/src/GCEED/common/sendrecv_copy.f90
+++ b/src/GCEED/common/sendrecv_copy.f90
@@ -21,8 +21,11 @@ integer :: ix,iy,iz,iob
 real(8) :: tpsi(mg_sta(1)-Nd:mg_end(1)+Nd+1,mg_sta(2)-Nd:mg_end(2)+Nd, &
                 mg_sta(3)-Nd:mg_end(3)+Nd,1:iobnum,1)
 
+!$omp parallel default(none) &
+!$omp          shared(iobnum,mg_sta,mg_end,tpsi,psi) &
+!$omp          private(iob,iz,iy,ix)
 do iob=1,iobnum
-!$OMP parallel do private(iz,iy,ix) 
+!$omp do collapse(2)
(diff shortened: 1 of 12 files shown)


mn2007:
Thank you for your implementation, @yhirokawa san. I'll take a benchmark and reply this PR later, and I'll consider your suggestion.

mn2007:
I confirmed that the implemented code is 1.34 times faster than before. It's too amazing for me. I'd like to pass this PR, but one problem is left. Fujitsu compiler doesn't pass the build for this code. I'll specify the point by a review function.

mn2007 (changes requested):
In src/GCEED/common/sendrecv_copy.f90, could you move two private(iz, iy, ix) sentences from inside an iob loop to outside the iob loop? In other word, could you write private(iob, iz, iy, ix) before the iob loop?

yhirokawa-ccs:
@mn2007 san, thank you for reviewing. I fixed omp directives in sendrecv_copy.f90.

mn2007 (approved):
I approve this PR.

mn2007:
Thank you for the modification and the large contribution, @yhirokawa san. I merge this PR.

PR #177, opened 2017-09-08
Add condition to k_shift_wf_last

uemoto1:
## About This Pull Request

This pull request contains the modification to avoid to call `k_shift_wf_last` when `projection_option == 'no'`). Originally, the above subroutine provides the `_last_bandmap.out` file and the projected overlaps at the end of the real-time calculation. In the ARTED mode, this calculation is always performed regardless of the setting. However, it consumes relativity large computation time when the system size and number of the electrons are large. Therefore, we prefer to skip this routine when `projection_option` is not specified, due to the convenience of the general users.

## Acknowledgement
- The author thanks @ayamada224 san for the bug report and fruitful comments.

## Command to Jenkins
- testmode = all

### DIFF
--- a/src/ARTED/control/control_sc.f90
+++ b/src/ARTED/control/control_sc.f90
@@ -420,7 +420,9 @@ subroutine tddft_sc
 !====Analyzing calculation====================
 
 !Adiabatic evolution
-  call k_shift_wf_last(Rion_update_rt,10,zu_t)
+  if (projection_option /= 'no') then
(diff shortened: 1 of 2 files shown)


uemoto1:
Sorry, I used inadequate reference value for the verification process in the regression test. The previous reference was obtained from the old calculation with a bit different condition. Here I retry this test with updateding the reference value. (The new value is consistent to the present inputfile in test suites.)

uemoto1:
test this please

PR #183, opened 2017-09-13
Workaround for parallelized BLAS routine in OpenMP parallelized loop.

yhirokawa-ccs:
For Fujitsu's parallelized BLAS routines, the application crashes when invoking a routine under OpenMP parallelized loop.
However, for many-core processors, outermost loop must be parallelized due to OpenMP overheads.

I added two implementation as a workaround.
I think that Fujitsu should be fixed it's problem.

### DIFF
--- a/src/ARTED/GS/diag.f90
+++ b/src/ARTED/GS/diag.f90
@@ -29,13 +29,18 @@ Subroutine diag_omp
   integer           :: ik,ib1,ib2
   integer           :: ia,j,i,ix,iy,iz,thr_id
   real(8)           :: kr
-  integer           :: ithr
 
 !LAPACK
   integer                :: lwork,info
-  complex(8),allocatable :: za(:,:),zutmp(:,:,:)
+  complex(8),allocatable :: za(:,:)
   complex(8),allocatable :: work_lp(:)
   real(8),allocatable    :: rwork(:),w(:)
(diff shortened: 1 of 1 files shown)


mn2007:
Thank you for implementing, @yhirokawa san. I'll check whether it successfully runs on a Fujitsu environment and OFP.

uemoto1:
Thank you @yhirokawa san and @mn2007 san. I also confirmed the program successfully reproduces the computation results in our workstation. I think this PR is available to merge to the developer branch.

mn2007:
Thank you for merging, @uemoto1 san. Though this PR had already merged, I report that sc and ms calulations in the ARTED part successfully finished on both a FX10 machine and the OFP even if I specified export FLIB_FASTOMP=TRUE on the FX10.

uemoto1:
Thank you for @mn2007 san for reporting and precise testing. I have to apologize that it is too early to merge before you have finished the checking process.

mn2007:
Well, thank you for saying that, @uemoto1 san. For this case, if there was a problem, we could give another PR. It may be difficult to deteremine when we should merge. I think it's better to wait to merge if other peaple leave comments.

uemoto1:
@mn2007 san. That is correct, just as you said. I will take care for it.

Issue #179, opened 2017-09-11
Bug of GS-SCF convergence in the current/recent version (ARTED side)


ayamada224:
I found a bug of ground-state scf calculation in ARTED side in the current/recent version (not in the older version: at least around two weeks ago).
(double check was done by Uemoto-san)

In the current version, the scf does not converge like (XX_GS.out):
 ...............
   148   -0.4379181722E+03    0.7835661139E-02    0.8154006743E-02    0.2362924103E-01
   149   -0.4380639074E+03    0.4877358546E-02    0.6855756716E-02    0.1908358073E-01
   150   -0.4379719259E+03    0.5001770535E-02    0.7146741877E-02    0.2145845483E-01

but on the other hand scf converges with the older version (of around the end of August):
.............
   148   -0.4384640347E+03    0.1445445105E-09    0.2410829159E-08    0.1922763162E-07
   149   -0.4384640347E+03    0.2467451177E-09    0.3026866230E-08    0.1054536409E-05
   150   -0.4384640347E+03    0.2446715902E-09    0.2720838887E-08    0.9017800164E-06

I'm going to fix this issue with Uemoto-san. 
If you know something about that please let us know.

best, 
A-Yamada

uemoto1:
Thank you @ayamada224 for bug report.

I also confirmed the above problem in recent version of SALMON.

Now I have submitted a job to search the first version causes this problem with rolling back the git histories. For now, I narrowed down the beginning of bug is one of pull request between PR#164-177.  I need a bit time to identify the single pull requests.

uemoto1:
I have obtained the result of calculations and identified the related pull request (PR #175). After merging this PR,  the decreasing of the convergency appears in the numerical results. I suggest checking the implementation of PR and (if it is necessary) reverting the merge commitment.

|Commit|Pull Request| Total energy (iter=100)| Delta E (iter=100)|
|:--|:--|:--|:--|
|3760346|Merge pull request #177|-31.1644588347157|-8.162118815171482E-004
|077efe5|Merge pull request #176|-31.1704499164816|-3.352016921326140E-003
|503b107|Merge pull request #175|-31.1695048636110|-1.105028257521923E-002
|e0f77a3|Merge pull request #173|-31.2658878785857|-1.240643143773923E-009
|409ffcb|Merge pull request #172|-31.2658878785857|-1.240643143773923E-009
|4a98673|Merge pull request #170|-31.2658878785857|-1.240643143773923E-009
[...]

PR #184, opened 2017-09-13
Modified 001_bulk_gs_rt

uemoto1:
## Overview
To avoid the simillar problem claimed in issue #179, I prefer to modify the verification condition of the regression test to more rigorous ones.

- Update the reference value of the regression test
- Narrow down the permissive error range
- Add the check of the convergence at the end of SCF loop
 
## Command to Jenkins
- testmode = all


uemoto1:
Thank you for merging, @yhirokawa san!

PR #185, opened 2017-09-14
use small capitals

mn2007:
According to coding rules, I do minimum replacement for variables whose first characters are large capitals. I didn't touch to src/ARTED, src/GCEED, examples/bulk_Si. If there are any requirements, I'll modify.


shunsuke-sato:
@mn2007 san, thank you for your implementation. I checked and merge it.

mn2007:
Thanks a lot for checking and merging, @shunsuke-sato san.

PR #188, opened 2017-09-19
Add input keyword opt md arted

ayamada224:
(1) I added new input keyword categories (see updated manual):
&opt
&md

(2) I made some changes in md option. Main change is to add "init_md" subroutin. 
Currently just generation of initial velocity with Maxwell-Bortzman distribution.
(then, I added "velocity" as global variable)

### DIFF
--- a/manual/input_variables.md
+++ b/manual/input_variables.md
@@ -18,6 +18,8 @@
 - [&analysis](#analysis)
 - [&hartree](#hartree)
 - [&ewald](#ewald)
+- [&opt](#opt)
+- [&md](#md)
 
 ## &calculation
 <dl>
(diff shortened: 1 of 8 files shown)


uemoto1 (approved):
Thank you @ayamada224 san, for the contribution and documentation.
I have checked this program successfully works in the KNL environment, and (at least) the ARTED level calculation reproduces the previous results. I confirmed that the modification of the new variables in the `src/io/inputoutput.f90` is implemented by appropriate method.

I think this PR is available to merge to the developer's branch,

As the future plan, I recommend to add the sample input (and regression tests) for this MD calculation.

Thank you!

PR #191, opened 2017-09-20
Update help messages of configure.py

yhirokawa-ccs:


uemoto1:
Thank you for developing @yhirokawa san, I will check it as soon as possible.
(Since the regression test succeeds after this modification, I guess this PR has no problem...)

uemoto1:
Okay, I confirmed the help message shown in `configure.py --help` is successfully changed.
I agree this PR is available to merge. Thank you for contribution, @yhirokawa san!

PR #192, opened 2017-09-25
check a calculated result for optical response

mn2007:
I modify testsuite to check a result of the optical response calculation for the C2H2 system. I checked python program works correctly, but I'm not sure if it also works on jenkins. @uemoto1 san, could you check this PR?


uemoto1:
Thank you @mn2007 san, I will review this PR.

uemoto1:
Thank you @mn2007 san for developement. I checked the result of the Jenkins and I confirmed there is no significant problems. I agree your code works well and available to merge the developer branch.

mn2007:
Thank you for checking and merging, @uemoto1 san.

PR #193, opened 2017-09-25
modify a frequency of a pulse.

mn2007:
I found that a calculated peak of oscillator strength for the isolated C2H2 molecule is not at 9.26eV, but at 9.28eV. So I modify a frequency of a pulse.


uemoto1:
Thank you @mn2007 san. I merge this PR.

PR #195, opened 2017-09-25
check a calculated result for a pulse

mn2007:
I prepare testsuite for a pulse calculation to check its result. @uemoto1 san, could you check again?


mn2007:
Thank you for your suggestion. I'm not familiar with python, so your suggestion is helpful for me.

uemoto1 (approved):
Thank you @mn2007 san. I think your code does not have a significant problem and works well.
I have one recommendation: the loop on the line 38 in `verification` is meaningless and I suggest it is convenient to remove for the future upgrades of `checklist`.

uemoto1:
I will merge this PR after finishing the present build test.

mn2007:

PR #196, opened 2017-09-26
Modify the dielectric function calculation

uemoto1:
## Overview
This pull request provides the improvements for the dielectric function calculation routines for the periodic system.
1. Added the exporter (`SYSNAME_lr.data`) of the dielectric function, which has been proposed at the June meeting. (https://www.dropbox.com/s/43qyhxj0n5y7a5z/uemoto0805_output_format.pdf?dl=0) Besides, the old output file (`SYSNAME_e.out`) is also remained for the compativility.

2. Implemented the zero frequency correction of conductivity calculation, which avoids the divergence of the dielectric function caused by the existence of the numerical error. This correction is automatically enabled in the case of the electronic temperature < 0  (insulator case).

dev_dielec

- Dielectric function of the silicon crystal with comparing between the original result (dotted curve) and modified method (solid curve). In the case of the modified code, it is obviously that the divergence at the low frequency region is successfully removed.

## Future work
- Presently, this implementation for the output file `SYSNAME_lr.data` only supports the atomic units. For the future work, the export in fs_ev units will be added in the output routine.

### DIFF
--- a/src/ARTED/RT/Fourier_tr.f90
+++ b/src/ARTED/RT/Fourier_tr.f90
@@ -86,16 +86,16 @@ Subroutine Fourier_tr
              &,(real(E_tot_w(ixyz)),ixyz=1,3)&
              &,(aimag(E_tot_w(ixyz)),ixyz=1,3)
       endif
+      
+      if (Trans_Longi == 'tr') then
+        write(fh_lr,'(1x,f13.7,12f22.14)') &
+            & hw &
+            &,(real(zsigma_w(ixyz)),ixyz=1,3)&
+            &,(aimag(zsigma_w(ixyz)),ixyz=1,3)&
+            &,(real(zeps(ixyz)),ixyz=1,3)&
+            &,(aimag(zeps(ixyz)),ixyz=1,3)
+      endif
     endif
-    
-    if (Trans_Longi == 'tr') then
-      write(fh_lr,'(1x,f13.7,12f22.14)') &
-          & hw &
-          &,(real(zsigma_w(ixyz)),ixyz=1,3)&
-          &,(aimag(zsigma_w(ixyz)),ixyz=1,3)&
-          &,(real(zeps(ixyz)),ixyz=1,3)&
-          &,(aimag(zeps(ixyz)),ixyz=1,3)
-    end if
   enddo
  
   if (comm_is_root(nproc_id_global)) then
--- a/src/ARTED/RT/Fourier_tr.f90
+++ b/src/ARTED/RT/Fourier_tr.f90
@@ -16,24 +16,43 @@
 !--------10--------20--------30--------40--------50--------60--------70--------80--------90--------100-------110-------120-------130
 Subroutine Fourier_tr
   use Global_Variables
+  use salmon_file, only: open_filehandle
   use salmon_parallel, only: nproc_id_global
   use salmon_communication, only: comm_is_root
   implicit none
   integer :: ihw,ixyz,iter
   real(8) :: hw,tt
   complex(8) :: jav_w(3),E_ext_w(3),E_tot_w(3)
+  complex(8) :: jav_d(3),jav_s(3),smt_s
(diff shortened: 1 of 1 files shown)


shunsuke-sato:
@uemoto1 san, thank you for your implementation.
I noticed that, in this routine, the time for the Fourier transform is wrongly shifted by `dt/2`. And thus, the newly added part also contains this wrong shift. I think we should fix it.

Related to this correction for the low frequency component, I came up with another way to correct the breakdown of the gauge invariance. Please have a look at the following branch:
[shunsuke-sato/SALMON/corr_gauge_inv](https://github.com/shunsuke-sato/SALMON/tree/corr_gauge_inv)
It would be more natural than the artificial subtraction of the zero frequency component. Also, this way is applicable for both metal and insulator. Probably, we would combine the both methods for a temporal implementation.

uemoto1:
Thank you @shunsuke-sato san. I fixed this PR based on your suggestion.

shunsuke-sato:
@uemoto1 san, thank you for fixing the code.
Could you remove a commit that was merged from my branch? I did not expect that the commit was merged to the master, but just intend to show how to make a correction for the gauge invariance.

Also, since I did not attend that meeting and do not really understand what was decided, could you ask someone else to review/merge this PR?

mn2007:
@uemoto1 san and @shunsuke-sato san, I guess this was not discussed in the meeting because I haven't heard this matter so far. If @uemoto1 san removes the commit proposed by @shunsuke-sato san or discussion between you two successfully ends, I can check whether dielectric function doesn't diverge and merge this PR.

uemoto1:
Okay, I excluded the previous merge. Here, it is working and possible to execute the dielectric function calculation. Thanks for comments, @shunsuke-sato and @mn2007 san.
For convenience, I uploaded the example input file to obtain the proposed data:
https://gist.github.com/uemoto1/d6ee7b29ba279d7bd8f87f66f51df636

mn2007:
Then, I check this PR. Please wait for a moment.

mn2007:
I checked Si_lr.data on FX10 with 1 node but variable which is correctly displayed is only propagation time and other values are zero or NaN. In addition, file "fort.8" is generated when I used 12 nodes. I'll examine why this happens but could you also check whether there is a mistake or not, @uemoto1 san?

uemoto1:
Thank you @mn2007 san for reporting, I will check the problem.

uemoto1:
Okey, I fixed the problem you mentioned. Presently the writing error for `SYSNAME_lr.data` is resolved. (And the problem seen in regression test is also resolved)

mn2007:
Thank you for fixing the problem, @uemoto1 san. I check again.

mn2007:
Thank you for the implementation, @uemoto1 . The problem about fort.8 is solved but the problem about zero and NaN values is not solved in my environment. I'll consider this afternoon.

uemoto1:
Thank you @mn2007 san. It seems to be a bit complicated problem. Currently I obtained the appropriate results in the KNL environment. May I know the example of the output file?

uemoto1:
@mn2007 san. I start to understand the problem. This routine cause the result when the pulse shape is not assigned `impulse`. One solution is not to generate the `_lr.data` file if the pulse shape is not `impulse` shape. I will send the bug fix later.

Sorry for checking so many times...

mn2007:
I've noticed that I did a pulse caluclation. Sorry for making you confused. Now I'm checking an impulse calculation.

mn2007:
I obtained same results as @uemoto1 san gave. So I merge this PR. @shunsuke-sato san, could you give another PR if you wish to implement the correlation for the gauge invariance? Thank you for your cooperation.

uemoto1:
Thank you for checking and merging, and I'm sorry for helping the debugging. Because I'm harried to implement before the release, I think my check/test on the local environment is insufficient at this time. I apologize for taking your time when you're busy.

PR #197, opened 2017-10-02
parallelization/tuning of PS-update subroutine

ayamada224:
Subroutine of preparing/updating pseudo-potential (prep_ps_periodic) was more tuned/parallelized(openmp).


uemoto1:
Thank you @ayamada224 san, for developing.
According to the regression test, the parallelized PS routine successfully reproduces the previous results in our platform. I think this PR works well and available to merge.

yhirokawa-ccs (approved):

PR #203, opened 2017-10-13
Added input keywords: read_initial_guess and modify_initial_guess opt…

ayamada224:
…ions (ARTED side)

(1) New input keywords: "read_initial_guess" and "modify_initial_guess" were added in &control block.
(1-A) read_initial_guess:
This is useful for continuing GS calculations restarting from the previously obtained wave function ("gs_wfn_k" directory) if the previous GS calculation was not converged. Also it can be used for geometry optimization starting from the converged GS wave function. 

(1-B) modify_initial_guess:
This keyword is to modify the initial guess wave function to be efficient GS convergence. It works only if the previous read_initial_guess=y, and currenty, modify_initial_guess = "copy_1stk_to_all" is implemented, meaning that the first k-point wave function data (in "gs_wfn_k" directory) is copied to the other all k-points as initial guess, where the first k-point wave function is already pre-calculated as gamma-point. 

(2) Minor improvement in optimization iteration:  just single line was inserted to be a little efficient way, and also printed iteration loop number was fixed.

### DIFF
--- a/manual/input_variables.md
+++ b/manual/input_variables.md
@@ -96,6 +96,18 @@ Default is the current directoy, <code>./</code>.
 <dd>Name of a filename for the restart calculation.
 </dd>
 
+<dt>read_initial_guess; <code>Character</code>; 3d</dt>
+<dd>
+Option to read wave function as initial guess (pre-calculated "gs_wfn_k" directory) if this is <code>y</code>.
+Default is <code>n</code>.
+</dd>
+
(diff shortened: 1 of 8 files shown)


uemoto1:
Thank you @ayamada224 san, for developement. I will check it later.

uemoto1 (approved):
Thank you @ayamada224 san. I checked the new input variable is successfully added and checked the reproduction of the previous results. I think it is available to merge into the develop branch!

PR #204, opened 2017-10-16
OpenMP parallelization by space grids and orbitals in ground-state calc (ARTED side)

ayamada224:
OpenMP parallelization by space grids and orbitals in ground-state calc (ARTED side)

OpenMP parallelization in ground-state calculation was revised. If the number of k-points in a node is larger than the number of orbital, openmp works as k-points parallelization as before (no change in the case), but if the orbital number is larger than the k-point number, openmp parallelization works for space grid points and orbitals loops. As for the calculation time, if gamma point calculation of large system for example, the calculation time of GS option was reduced to 1/10 on OFP (256 threads in a node) (....I think it can be faster if further tuning is done in future). 
The revised subroutines are
(1) Gram_Schmidt
      Referring algorithm developed in RSDFT, openmp parallelization by orbital and space grid was implemented. (but not perfectly the same as RSDFT to avoid complication, so efficiency is not as good as RSDFT)
(2) CG
      Following Iwata-san's comment, Gram-Schmidt orthogonalization in the CG loop is removed. Then, openmp with orbitals are introduced. (orthogonalization is done before and after CG)
(3) diag
      Openmp by orbitals are introduced.

### DIFF
--- a/manual/input_variables.md
+++ b/manual/input_variables.md
@@ -541,6 +541,24 @@ Threshold for convergence check that is used when either <code>'pot'</code> or <
 Default is <code>-1d0</code> a.u. (1 a.u.= 1.10d2 Å<sup>3</sup>eV<sup>2</sup>)
 </dd>
 
+<dt>omp_loop; <code>Character</code>; 3d</dt>
+<dd>
+Loop for OpenMP parallelization if periodic boundary system is used. 
+<li>
+<code>k</code>: parallelization by k-point loop (Default).
+<li>
+</li>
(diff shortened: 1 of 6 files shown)


uemoto1:
Thank you @ayamada224 san, we will check it later.

syamada0 (approved):

syamada0:
Thank you for your implementation, @ayamada224 san.
I checked regression of the results and confirmed that the calculation cost is substantially reduced for [ nstate >> num_kgrid ] cases.  So I approve this pull request.

uemoto1 (approved):
Thank you for developing and reviewing, @ayamada224 san and @syamada0 san. I also confirmed the speedup of GS calculation and reproduction. The result of total energy is permittable with  using the modified CG routine. I think it is available to merge.

PR #205, opened 2017-10-18
Fixed a testsuite error when target architecture is specified.

yhirokawa-ccs:


uemoto1:
Thank you for contribution, @yhirokawa san. I will check the point of this PR.

uemoto1:
At least in our test environment (intel KNL platform), this modified CMake test is working as well as the previous case. Besides, I think your modification (automatic detection of MPI) is reasonable, and agree this is worth to merge the developer branch.

Thank you for your contribution, @yhirokawa san!

PR #207, opened 2017-10-20
Gs threshold arted

ayamada224:
(1) Convergence threshold of GS calculation was applied to ARTED side. The keyword "convergence" already prepared by GCEED side got to be available now in ARTED side. Then, new option, convergence="rho_dne" was introduced (as default: density difference divided by electron number, same as Octopas software). After this merge, the "rho_dne" option is going to be implemented also in GCEED side. (These changes are based on discussion with Noda-san and IMS members). 
The other small changes are
(2) addition of subroutine to check bad/unavailable input keywords/options combination. Currently only one thing is added. I recommend developers to add the bad input combinations to be user friendly software.
(3) tiny change of format of standard log to be better looking.

### DIFF
--- a/manual/input_variables.md
+++ b/manual/input_variables.md
@@ -521,22 +521,26 @@ subspace diagonalization during scf cycle.
 <dt>convergence; <code>Character</code>; 0d</dt>
 <dd>
 Choice of quantity that is used for convergence check in a scf calculation. 
-Default is <code>'rho_dng'</code>. The following can be chosen.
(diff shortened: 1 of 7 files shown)


syamada0 (approved):

syamada0:
Thank you for your contribution, @ayamada224 san.
I checked that the "rho_dne" mode works properly and reproduces previous results in ARTED part.
So I approve this pull request.

PR #208, opened 2017-10-20
implement rho_dne

mn2007:
I implement "rho_dne" in the GCEED part. To distinguish whether a threshold value for convergence is square of absolute values or square of norm, I add "norm_" to previously defined values for a variable "convergence". In addition, since I had noticed that we need not multiply subtle volume to norm, I remove the subtle volume for related parts. Moreover, I had to modify ARTED part related to a backup routine. 

I guess automatic tests don't pass because of shutdown of Jenkins. But I checked that the run for both GCEED part and ARTED part successfully finshed. 

Conversations are mainly done with @ayamada224 san, so I ask @ayamada224 san to review. But if there are any opinions to this PR, please comment.


mn2007:
I found mistakes in this implementation. Could you wait to review till I commit the modification, @ayamada224 san? Sorry for that.

ayamada224:
Thank you @mn2007 san, I'll review the code after you fixed the mistakes. It may take time to review but I'll try it by Monday.

mn2007:
Thank you so much for your reply, @ayamada224 san. I've finished the modification. I'll wait by Monday.

PR #209, opened 2017-10-24
Fixed a bug in compiling the stencil computation.

yhirokawa-ccs:
In ARTED part, the stencil computation always uses compiler vectorization (Fortran90) codes regardless the build configuration.
Maybe, this bug was caused by my previous PR.
I apologize this problem, and fix it by this PR.


uemoto1:
Thank you @yhirokawa san, I will check it later.

PR #211, opened 2017-10-27
Optimized non-local pseudopotential operations

shunsuke-sato:
I optimized the nonlocal pseudopotential operations by introducing the projector array `zproj` and reducing the number of operations. I also gathered the scattered phase updates into `src/ARTED/common/projector.f90` to simplify the program.

For `examples/bulk_Si/input_sc_Si.inp` and `examples/bulk_Si/input_sc_Si.inp`, 
- the pseudopotential operation gets 20-30% speedup
- the hamiltonian operation gets 8-13% speedup
- the total dynamics time gets 4-9% speedup.

@yhirokawa san, could you have a look at this patch, please? I'm a bit afraid that I broke the timer in the code.

### DIFF
--- a/src/ARTED/GS/CG.f90
+++ b/src/ARTED/GS/CG.f90
@@ -37,6 +37,7 @@ Subroutine CG_ompk(iter_cg_max)
   use salmon_communication, only: comm_summation
   use timer
   use hpsi, only: hpsi_omp_KB_GS
+  use projector
   implicit none
   real(8),parameter :: delta_cg=1.d-15
   integer iter,ik,ib,ibt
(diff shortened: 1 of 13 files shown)


yhirokawa-ccs (approved):
A subroutine ``update_projector`` does not support OpenACC, but currently SALMON can not be performed on GPU.
I will merge it after checking the behavior.

yhirokawa-ccs:
I checked the performance under 16 KNLs for ``examples/bulk_Si/input_sc_Si.dat``.
I seem that the timer hasn't any problem. I will merge it.

1. the performance of pseudopotential gets up to approximately 7-9%.
2. the performance of hamiltonian gets up to approximately 2-4%.

shunsuke-sato:
@yhirokawa san, thank you for quick reviewing and merging. As you mentioned, `update_projector` does not support OpenACC yet since I do not have the GPU environment and could not check the implementation.
Perhaps, we need the GPU environment for SALMON...

PR #213, opened 2017-10-31
Bugfix of Multiscale Calculation

uemoto1:
## Overview
This pull request contains some improvements and bugfixes of the 1-3D multiscale calculation. Besides, It also appends a few optional input parameters. Since the backward compatibility remains in this code, the input file of the multiscale calculation in the previous version is still available in the modified code.

## Changes
- Enable the arbitrarily shaped structure of the macro points.
- Enable `3D` mode of the multiscale calculation in SALMON. 
(Previously these two points are implemented only in the ARTED)
- Automatic detection of `NKSplit` and `NXYSplit` parameter.
(Currently, these parameters remain as the optional items for the compatibility of the old input file. We have a plan to remove these variables in the future updates)
- Shift `1/2 Delta t` in the numerical integration of the energy absorption. It is not significant to results, but in principle, it is better for the energy conservation.
- Change the log format in the STDOUT of the multiscale calculation. The new format is easier for users to grasp how the computation proceeds.
- Additionally, insert the OpenMP directives to speed up the data assignments.
- Fix any other minor problems...

### Comparison between original results and modified code
- As shown in the following figure, the new code successfully reproduces the previous ones:
fig_result1

### Example of the output (for peer reviewers)
- The input file uploaded as below:
https://gist.github.com/uemoto1/e90ce9f3c015c086c6b377b83bc5c213
( The pseudopotential `Si_rps.dat` is also required for execution)

- You can get the results at iter=1000 in STDOUT
```
 -----------------------------------------------
 Multiscale iter =  1000/  1000
 Microscopic system:
    Excitation energy:     Eex =  6.533425E-003    Diff =  2.233543E-003 [au]
 Macroscopic system:
      Absorbed energy:    Eabs =  6.531510E-003    Diff =  2.233342E-003 [au]
         Field energy:    Eemf =  2.749756E+000    Diff = -2.429957E-003 [au]
      Total EM energy:    Etot =  2.756287E+000    Diff = -1.966158E-004 [au]
 Position of wavepacket:
     Central position:     <r> = -2.035E+004,  0.000E+000,  0.000E+000, [au]
    Spatial spreading:     rms =  8.669E+003,  0.000E+000,  0.000E+000, [au]
 Total time =   33.6190788745880     
 -----------------------------------------------
```

I have tested this code in the Oakforest PACS environment. Please someone check in FX machine?

Thank you for your consideration.


uemoto1:
I found a problem in the export of `Ac_m` file in some condition. I send bugfix later.

uemoto1:
Currently I send the bugfix. Now The code is working correctly. (I guess I make mistakes to resolving the conflicts...?)

shunsuke-sato:
@uemoto1 san, thank you for your implementation. Unfortunately, I do not have an access to FX machine. So, I would like to ask someone to test this patch.

I have one suggestion about absorbed energy. In the present implementation, the absorbed energy is computed by Ohm's law. However, based on my experience, the divergence of Poynting vector would be more numerically accurate (please have a look at [PRB 92, 205413 (2015)](http://journals.aps.org/prb/abstract/10.1103/PhysRevB.92.205413) for multi-scale simulations. So, please consider to employ Poynting vector if you suffer from the numerical accuracy.

uemoto1:
@shunsuke-sato san, Thank you for comments. I think it is possible to add the Poynting vector routine. However, I guess the coding and the check of precision will be a time consuming. I will schedule it for the next implementation. 
(P.S. Since I got some comments for the preparation of the next release version, I have to modify this PR before merging. After modification, I will ask the check and merge. Sorry for inconvenience...)

src/ARTED/control/control_ms.f90
```diff
@@ -104,7 +107,7 @@ subroutine tddft_maxwell_ms
   write(file_energy_transfer, "(A,'energy-transfer.out')") trim(directory)
   write(file_ac_vac, "(A,'Ac_Vac.out')") trim(directory)
   write(file_ac_vac_back, "(A,'Ac_Vac_back.out')") trim(directory)
-  write(file_ac_m, "(A,'Ac_M',I6.6,'.out')") trim(process_directory), NXY_s
+  write(file_ac_ms, "(A,'Ac_m',I6.6,'.out')") trim(process_directory), imacro
```
shunsuke-sato: Here, 'imacro' becomes 'nmacro_e+1' because of the above loop at line 87. But, I think this behavior is not expected. Please check it.

shunsuke-sato (commented):
@uemoto1 san, thank you for your implementation. I tried to review and check this PR. However, the file of `Ac_m****` are not properly created. I added a comment on the code. Please have a look at it.

uemoto1:
@shunsuke-sato san, Thank you for checking. I will send the bugfix later.

uemoto1:
Thank you for many fruitful comments! Now, I have fixed related problems and reduced warning messages.

shunsuke-sato:
@uemoto1 san, thank you for fixing the problem. However, it seems that this patch does not reproduce the original results. For example, if `examples/bulk_Si/input_ms_Si.inp` is used, `Ac_M000000.out` of the original code and `Si_Ac_m_000001.out` of this patch are not identical.
Furthermore, all the files of `Si_Ac_m_00000*.out` are identical although they should give the data for different macro points.

So, could you check the output of this patch?

uemoto1:
@shunsuke-sato san, At least, in my environment, the code generates the appropriate output data; I can't reproduce the bug report. However, during checking the source codes (the related section of  Ac_M outputs), I found one subroutine accesses to an array element which placed on out of the allocated range. I guess this buffer-overflow problem will probably induce the similar problem you reported by contaminating the stored data. I immediately send a bugfix. In my test, it also works correctly as below:
compare1

 (In this time, I increased NX_m=24 to clarify the error the induced by growing up the propagation thickness. The input file is uploaded as above comments.)

I think this behavior strongly depends on the system size, processors number,  compilers, and any other environments. Could you check the modified code and reproduction of the error in your environment?

Thank you for your corporation.

shunsuke-sato:
@uemoto1 san, still this patch does not reproduce the original results, at least, in my environment. In the attached figure, the red line shows the original result, and the blue-dotted line shows the result of this patch. You can see the file information from the legends of the figure.

Do you get the identical lines for the vector potential in time?

pr213

uemoto1:
Thank you @shunsuke-sato for reporting. I have recalculated by using "example/bulk_si" inputs, and the attached is the result of the calculation.

The blue line is the original results and orange point are the modified code. For this comparison, I use the version 176863c59f5ec498741985361772c68bc0121db1 as old version.

uemoto1:
@shunsuke-sato san, During testing in some of different computation conditions, I confirmed an appearance of significant error will be seen when the number of MPI are gradually increased (much larger than number of the macropoints). At that time, the behavior of the differences are resemble to the attached in your report.

uemoto1:
I have experimentally fixed the bug as above.

shunsuke-sato:
@uemoto1 san, thank you for the confirmation.
In my calculations, I used 8 MPI processes for 4 macroscopic points. And, this is the only possible choice for the original code with `examples/bulk_Si/input_ms_Si.inp`.

uemoto1:
Thanks for checking @shunsuke-sato san. I grasp the bugfix proposed above is also works successfully for the MPI number of nproc = 8.

uemoto1:
The modification provide the header comment lines of the output files which helps for users to indicate the meaning of the columns and its unit. It also change the description of term `Ac_x` to `A_x/c` in the comment line for convenience. For example ...
```
# Time[au] A_x/c(0)[au] A_y/c(0)[au] A_z/c(0)[au] A_x/c(L)[au] A_y/c(L)[au] A_z/c(L)[au]
```
And change and add the defination of detection coordinates.

shunsuke-sato:
@uemoto1 san, thank you for fixing the problem. Then, I could restart the review and the tests. But, I could not have the time until Friday because of visitors. So, could you wait until Friday? I'm a bit afraid that the first release is urgent. But, I do not know the exact plan.

Also, as I wrote above, I do not have access to FX machines. So, could you ask someone else for the tests on FX machines?

mn2007:
I'll try the calculation on the FX machine this afternoon, so could you wait for a while? I'll give a comment after the calculation.

mn2007:
I found two problems.

1) The build on the k-computer did not succeed. The build passes If line 660 in src/ARTED/control/control_ms.f90 is modified from
<pre>
!$omp parallel do default(shared) collapse(3) &
!$ & private(iix_m, iiy_m, iiz_m, rrx, rry, rrz, eem) &
!$ & reduction(+: sx1, sy1, sz1, sx2, sy2, sz2, sem)
</pre>
to 
<pre>
!$omp parallel do default(shared) collapse(3) &
!$omp private(iix_m, iiy_m, iiz_m, rrx, rry, rrz, eem) &
!$omp reduction(+: sx1, sy1, sz1, sx2, sy2, sz2, sem)
</pre>
@uemoto1 san, could you modify this part? 

2. Even if the problem gave above is solved, the result seems to be strange. I attach the calculated result at iter=4000.
em
Is it possible for you to fix?

uemoto1:
Thank you @mn2007 san. Your report is very helpful!
1.  I will fix the directives of the OpenMP.

2. I think the difference is originating to the new output file uses the different column arrangement (and unit system) from the old ones. Due to this reason, I think the direct comparison between old output and new output is a bit difficult.

Currently, I have the series of output files tested in the KNL machine (We have already checked it is equivalent to the older results). Later, I will upload one file of KNL output. Can you compare with the K-computer's result to the KNL result?

mn2007:
Thank you for your proposal and sorry for skipping over the description for the output, @uemoto1 san. It will take a time to understand new format, so uploading the KNL result would be very helpful for me. I'll wait until you upload.

uemoto1:
Sorry for late, @mn2007  san. I fixed the OpenMP directives at this commitment.

The attached is the example of the input file as below:
https://www.dropbox.com/s/emb8ozlvh4n5rdx/input_ms_Si_check1d.txt?dl=0
(Please prepare `Si_rps.dat` file on the same directory of input file) 

The following is the example of output file at `iter = 4000` step (obtained by KNL):
https://www.dropbox.com/s/fsbx7pkdqvgx8i9/ms_Si_check1d_Ac_004000.out?dl=0

The critical quantities in this file is 
- Column 1: Horizontal Axis
- Column 16: Energy of Electronic System
- Column 17: Energy of Absorption

fig_test

The 16-th and 17-the elements are calculated from the most of other physical quantities in the multiscale calculation. Therefore, we can use those couple of values as the representation of the convergence.

mn2007:
Thank you for updating, @uemoto1 san. But it seems that problems are not solved yet. 

1. The part where build does not pass is what I commented. But updated parts are not there. Therefore please let me explain in other words. Because `!$` is used instead of `!$omp` on lines 661 and 662 in src/ARTED/control/control_ms.f90, the build doesn't pass. Could you modify this part? 

2. Calculated values on the k-computer are a little bit different from the ones on OFP though I used the uploaded input file. Is this differece acceptable or not? 
em

uemoto1:
Sorry @mn2007 san, I have misunderstood your point. I retried to fix the OpenMP directives.
Besides the small difference between OFP and K is I think nearly acceptable...

mn2007:
Thank you for fixing the problem about the OpenMP directive, @uemoto1 san. Since @uemoto1 san insisted that the difference is acceptable, I merge this PR.

uemoto1:
Thank you @mn2007 san for quick checking...

Generally, the edge point the error induced by the numerical differentiation will be more easy to see than any other point, due to the discontibuity of the current density. Therefore,in  this case, I guess the such kind of the errors grows up by the difference of the processer and paralization method,

(Besides, I want to grasp the actual numerical precision. Can I have the output files  of `ms_Si_check1d_Ac_M_0000??.out` in the K computer ?. At least, the beginning three files such as `M_000001`, `M_000002` and `M_000003`?

mn2007:
Followings are files obtained on the K computer. Could you tell me if you can not download the files? 

https://www.dropbox.com/s/zu9mwivdjr83geg/Si_Ac_M_000001.out?dl=0
https://www.dropbox.com/s/dwv1zd8a863your/Si_Ac_M_000002.out?dl=0
https://www.dropbox.com/s/66a4rrgds91njdd/Si_Ac_M_000003.out?dl=0
https://www.dropbox.com/s/8kvkzqf7eqky3h6/Si_Ac_M_000004.out?dl=0

uemoto1:
Thank you @mn2007 san, I can download your data. They are helpful for my analysis!

mn2007:
Could you wait your analysis? I uploaded wrong output files. Sorry for that.

uemoto1:
Okay!, Please tell me when you finished the replacement of these files.

mn2007:
Sorry for making you confused. I uploaded correct ones. Could you check for these files?

https://www.dropbox.com/s/2gjix3ywrbhnd8n/ms_Si_check1d_Ac_M_000001.out?dl=0
https://www.dropbox.com/s/vkj6suce4p4op94/ms_Si_check1d_Ac_M_000002.out?dl=0
https://www.dropbox.com/s/ah1skcti720iqwn/ms_Si_check1d_Ac_M_000003.out?dl=0
https://www.dropbox.com/s/i583ddszxx5e0ex/ms_Si_check1d_Ac_M_000004.out?dl=0
https://www.dropbox.com/s/oeqodokxuffxs7i/ms_Si_check1d_Ac_M_000021.out?dl=0
https://www.dropbox.com/s/bx8rww6ihxfir7l/ms_Si_check1d_Ac_M_000022.out?dl=0
https://www.dropbox.com/s/p2ymyu4w635nrl2/ms_Si_check1d_Ac_M_000023.out?dl=0
https://www.dropbox.com/s/1vmwz00hamz1v05/ms_Si_check1d_Ac_M_000024.out?dl=0

uemoto1:
Thank you for your reply and sending items despite being so busy!

PR #215, opened 2017-10-31
Reduced memory pressure when updating non-local pseudopotential.

yhirokawa-ccs:
I optimized non-local pseudopotential update from the point of view of a memory pressure.
They pack (gather) pseudopotential vector before updating a domain. After computing, they unpack (scatter) to 3D space grid.

I check the performance with 16 KNL nodes and ``examples/bulk_Si/input_sc_Si.inp``

1. the performance of pseudopotential improves up to 40%
2. the performance of hamiltonian improves up to 12%

### DIFF
--- a/src/ARTED/common/hpsi.f90
+++ b/src/ARTED/common/hpsi.f90
@@ -97,8 +97,13 @@ contains
 
   subroutine hpsi_omp_KB_base(ik,tpsi,htpsi,ttpsi)
     use timer
-    use Global_Variables, only: NLx,NLy,NLz,kAc,lapx,lapy,lapz,nabx,naby,nabz,Vloc,Mps,iuV,Hxyz,Nlma,a_tbl,zproj
+    use Global_Variables, only: NLx,NLy,NLz,kAc,lapx,lapy,lapz,nabx,naby,nabz,Vloc,Mps,iuV,Hxyz,Nlma,zproj
     use opt_variables, only: lapt,PNLx,PNLy,PNLz,PNL
+#ifdef ARTED_STENCIL_PADDING
+      use opt_variables, only: zJxyz => zKxyz
+#else
+      use opt_variables, only: zJxyz
+#endif
 #ifdef ARTED_USE_NVTX
     use nvtx
 #endif
@@ -126,7 +131,7 @@ contains
     LOG_END(LOG_HPSI_STENCIL)
 
     LOG_BEG(LOG_HPSI_PSEUDO)
-      call pseudo_pt(ik,tpsi,htpsi)
+      call pseudo_pt(ik,zJxyz,zproj(:,:,ik),tpsi,htpsi)
     LOG_END(LOG_HPSI_PSEUDO)
 
     NVTX_END()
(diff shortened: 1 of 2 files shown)


src/ARTED/common/hpsi.f90
```diff
@@ -150,34 +155,54 @@ subroutine subtraction(Vloc,tpsi,htpsi,ttpsi)
...
-        ia=a_tbl(ilma)
-        uVpsi=0.d0
+      complex(8) :: pseudo(NPI)
+      complex(8) :: dpseudo(NPI)
+
+      dpseudo = cmplx(0.d0)
+
+      ! gather (load) pseudo potential point
+      do i=1,NPI
+        pseudo(i) = tpsi(idx_proj(i))
```
shunsuke-sato, line 173: I think that some parts of `idx_proj(i)` are indefinite since the first dimension of `zJxyz(:,:)` is allocated up to `Nps` but initialized only up to `Mps(ia)` for each ion. Thus, if `Mps(ia)` has different value for different ions, the use of the full array `idx_proj(1:NPI=Nps*NI)` may induce segmentation fault.

src/ARTED/common/hpsi.f90
```diff
@@ -150,34 +155,54 @@ subroutine subtraction(Vloc,tpsi,htpsi,ttpsi)
...
         do j=1,Mps(ia)
-          i=zJxyz(j,ia)
-          htpsi(i)=htpsi(i)+zproj(j,ilma,ik)*uVpsi
+          dpseudo(ioffset+j) = dpseudo(ioffset+j) + zproj(j,i) * uVpsi
         end do
       end do
+      end do
+
+      ! scatter (store) pseudo potential point
+      do i=1,NPI
```
shunsuke-sato, line 197: The use of the full array `idx_proj` may induce segmentation fault. Please have a look at the above comment.

src/ARTED/modules/opt_variables.f90
```diff
@@ -204,6 +210,21 @@ subroutine init_for_padding
     end do
   end subroutine
 
+  subroutine init_projector
+    use global_variables
+    implicit none
+    integer :: i
+
+    NP = Nlma / NI
+    NPI = Nps * NI
```
shunsuke-sato: Related to above comments, `NPI` might be efficiently described as `NPI=sum(Mps)`.

shunsuke-sato (commented):
@yhirokawa san, thank you for your implementation. The optimization is very useful for all of users.

I had a quick review and have a couple of comments.
Perhaps, I misunderstood something. In that case, please let me know.

src/ARTED/common/hpsi.f90
yhirokawa-ccs: Thank you for reviewing.
I applied your suggestion by the commit ade3d81 that this commit removes uninitialized elements.

shunsuke-sato:
@yhirokawa san, thank you for quick implementation. But, still the problem does not seem to be solved. I check SiO2 calculation, but this patch does not reproduce the previous result.

Since I know how to fix the problem, let me fix it first. Then, I'll send a pull-request to your branch.

yhirokawa-ccs:
@shunsuke-sato san, thank you for quick check.
I found the cause of bug at SiO2 calculation (different value for different ions).
I will send fix commit tomorrow.

shunsuke-sato:
@yhirokawa san, thank you for your response.
Basically, what I wanted mention is that bug. Sorry, my explanation was not so clear.

Then, I'll wait for the fix commit. Thank you for your implementation.

yhirokawa-ccs:
@shunsuke-sato san, I apologize that the bug. I think that the bug fixed.
However, I haven't SiO2 calculation input file (Currently, ``examples`` and ``testsuites`` support Si case only...), so I cannot check the result under the environment of your reported.
We should check the result of SiO2 at tests.

src/ARTED/modules/opt_variables.f90
```diff
@@ -204,6 +212,27 @@ subroutine init_for_padding
     end do
   end subroutine
 
+  subroutine init_projector(zJxyz)
+    use global_variables
+    implicit none
+    integer,intent(in) :: zJxyz(Nps,NI)
+    integer :: i,ioffset
+
+    NP = Nlma / NI
```
shunsuke-sato: The number of projectors may be different for different ions. Thus, `NP` can be different from `Nlma/NI`.

shunsuke-sato (commented):
@yhirokawa san, thank you for your implementation. But, I found another problem. See the comment on the code.

This pseudopotential part is one of the most complicated part of the code and based much on physical intuitions. So, I'll make a fix commit and check it by myself. Then, I'll make a pull-request to your branch `pseudo-pt`.

Unfortunately, I could not manage the time for this task this week. Most probably, I'll send the pull-request to your branch on Sunday. So, I would like to ask you to wait for the fix commit. Once you will merge the fix commit to this branch, I'll merge this PR to `develop`. I'm sorry for inconvenience.

yhirokawa-ccs:
@shunsuke-sato san, I fixed it with the we count the number of projector as following codes.

```fortran
function count_if_integer(vec, val) result(n)
  implicit none
  integer,intent(in) :: vec(:)
  integer,intent(in) :: val
  integer :: n, i
  n = 0
  do i=1,size(vec)
    if (vec(i) == val) then
      n = n + 1
    end if
  end do
end function

allocate(nprojector(NI))
do i=1,NI
  nprojector(i) = count_if_integer(a_tbl, i)
end do
```

shunsuke-sato:
@yhirokawa san, thank you for fixing the problem! I checked the implementation, and confirmed that this patch reproduces the original results!
I also found about 7% speedup of the pseudopotential operation for `examples/bulk_Si/input_sc_Si.inp` with the Intel Xeon CPU!

PR #216, opened 2017-11-01
change name of array for pseudopotential

mn2007:
This PR is based on @syamada0 san's request. 

I change name of array for pseudopotential. For variables about pseudoptentails, I put suffix "_all" to the variables used in a whole region, and remove suffix "2nd" from the variables used in parallelized regions. I also modify src/GCEED/module/hpsi2_test.f90. I confirmed that results don't change even if an option HPSI_TEST=ON is used. 

In addition, I also remove the part related to iwk_size=11 or 12 because this part is not used at the moment. 

@syamada0 san, could you check this PR?


syamada0:
Thank you @mn2007 san, I will check it later.

syamada0 (approved):

syamada0:
@mn2007 san, thank you so much for responding to my request.
I checked the code and confirmed that there was no problem.

mn2007:
Thank you for checking and merging, @syamada0 san.

PR #220, opened 2017-11-08
modify filename for analysis 

mn2007:
I modify filename for analysis in a real time calculation. By this modification, the filename become name such as dns_??????.cube and it is arranged in accending order in iteration of time. In addition, I fix a bug for a problem that ELF is not calculated in the real time calculation. 

@uemoto1 san, could you check this PR?


uemoto1:
Thank you @mn2007 san, I will check it later...

uemoto1:
Thank you @mn2007 for implementation, I grasped the series of the output file is obtained.

mn2007:

PR #222, opened 2017-11-08
reimplement openmp in pseudopotential calculation

mn2007:
In PR#216, total elapsed time of an optical response calculation in the isolated system slowed down from `165.2` sec to `169.7` sec with 12 nodes on the K computer . By this implementation, it recovers to `164.0` sec.

@syamada0 san, could you check this PR?


syamada0:
Thank you @mn2007 san. I will check it later.

syamada0:
@mn2007 san, I checked the performance of the code but could not confirm such speedup on the Oakforest-PACS computer.
However, in principles, this modification cannot negatively effect the performance.
So I merge this PR.

mn2007:
Thank you for checking details and merging, @syamada0 san.

PR #224, opened 2017-11-09
Remove the parameters from examples/bulk_si

uemoto1:
I removed the symmetry parameters from the example input file `input_si_(sc|lr|ms).inp` to avoid the missleading of users (which is discussed in the previous meeting). And removed `nksplit` and `nxysplit` which is useless in the present version.

### DIFF
--- a/examples/bulk_Si/input_lr_Si.inp
+++ b/examples/bulk_Si/input_lr_Si.inp
@@ -9,8 +9,6 @@
 &system
   iperiodic = 3
   al = 10.26d0,10.26d0,10.26d0
-  isym = 8 
-  crystal_structure = 'diamond'
   nstate = 32
   nelec = 32
   nelem = 1
--- a/examples/bulk_Si/input_lr_Si.inp
+++ b/examples/bulk_Si/input_lr_Si.inp
@@ -23,7 +23,7 @@
 /
 
 &functional
-  xc ='BJ_PW'
+  xc ='PZ'
   cval = 1d0
 /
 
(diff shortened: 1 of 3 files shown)


mn2007:
I'm checking but a lr calculation is so heavy. It doesn't end within 30 minites even if I use 512 nodes on the K computer. Could you wait for a moment, @uemoto1 san?

uemoto1:
Thank you for comment @mn2007 . I also feel the calculation without using the symmetry is a bit heavy to use in the example calculation...  I have couple of the plans for the symmetry parameters in the examples:
1. Remove `isym` in the sample with reducing the R/K grid points.
2. Keep  `isym` in the sample but does not explain this in the web tutorials.

I will discuss this point with Prof. @yabana. Please wait a moment.

mn2007:
Thank you for considering the problem, @uemoto1 san. I wait until you decide.

uemoto1:
@mn2007 san, Accordings to the discussion, we will change the exchange correlation and Nk parameter to speed up. I will send the bug fix later.

mn2007:
I'll try with these input files. Please wait for a moment.

mn2007:
Both calculations ended within 30 minites with 64 nodes on the K computer. (A multi-scale calculation ends within 30 minites with 8 nodes.) And calculated results are the ones which are supposed to be. So I merge this PR. Thank you for preparing files, @uemoto1 san.

uemoto1:
Thank you @mn2007 san, for quick checking and merging!

PR #227, opened 2017-11-10
Modified Filename

uemoto1:
This pull request changes the extension of output filename from `.out` to `.data`.


mn2007:
I confirmed that files with extention ".out" are no longer generated in lr/sc/ms calculation. So I merge this PR. Thank you for implementing, @uemoto1 san.

Issue #233, opened 2017-11-13
Failure of the restart function


shunsuke-sato:
After PR #221 , the restart function started failing.
@ayamada224 san, could you fix the problem?

ayamada224:
Thanks for pointing out, @shunsuke-sato san, I'll fix it.

ayamada224:
I reproduced SEGMENTATION fault in restarting. 
I checked with compiler option of "-CB -traceback", then, it found some access violations. These are not my changes, so they should be fixed by the one who changed them.

(1) line 254 in module/opt_variables.f90
      idx_proj(ioffset+1:ioffset+1+Mps(i)) = zJxyz(1:Mps(i),i)
I think, it should be like
      idx_proj(ioffset+1:ioffset+Mps(i)) = zJxyz(1:Mps(i),i)
(probably the second "+1" should be removed, but not sure)
This part was recently changed by @yhirokawa san at Oct,31

@yhirokawa san, could you please fix this part?


(2) after line 757 in common/restart.f90
    ! data_local_Ac                                                                
    mt = min(Nt, ubound(data_local_Ac,3))
    allocate(tmp3(3,nmacro_s:nmacro_e,0:Nt))
    tmp3(:,:,:) = 0.d0
    tmp3(:,:,0:mt) = data_local_Ac(:,:,0:mt)
More exactly, error was at "tmp3(:,:,0:mt) = data_local_Ac(:,:,0:mt)"
The allocation of data_local_Ac was changed recently by @uemoto1  san at Oct,24.
I think something is wrong in the recent change especially nmacro_s,nmacro_e??.
Similar treatments are done after this part. 

@uemoto1 san, could you please fix this part?

shunsuke-sato:
@ayamada224 san, thank you for clarifying the problem, and sorry for wrongly assigning you. 
When I tested SALMON, the problems turned out after the PR #221, but they already exist before it.

uemoto1:
By mistake, I have commented the following text on the page of different pull request, I rewrite it here for convenience.

> @ayamada224 san, Thank you for merging.
> I have struggled at the printf debugging, and I found something is strange in `read_external_input_for_restart` subroutine . In the case `input_for_restart` is not exist, only the root process skips the reading and broadcasting, but any other processes can't skip the broadcasts. > I think this mismatch leads the desync of MPIs...

I will send an pull request of this bugfix later...

PR #223, opened 2017-11-09
generate vtk files

mn2007:
vtk files are generated in the GCEED part by this modification. It's much easier to make movies by vtk files when we use paraview. @uemoto1 san, I guess you are so busy. Is it possible for you to check this PR? If not, I'll ask other people.

PR #234, opened 2017-11-14
Bugfix of the segmentation fault in the restart with resizing.

uemoto1:
This pull request contains a bugfix of Segmentation Fault in `restart.f90` claimed in Issue #233.

### DIFF
--- a/src/ARTED/common/restart.f90
+++ b/src/ARTED/common/restart.f90
@@ -755,6 +755,8 @@ contains
     deallocate(tmp2)
 
     ! data_local_Ac
+    if (use_ms_maxwell == 'y') then
+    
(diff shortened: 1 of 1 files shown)


uemoto1:
This PR is effective for the (at least) segmentation fault claimed in PR #223, However, I guess the origin of problem issued in #223 is not only caused by the #234 and #235. I think the additional bug tracking is necessary...

ayamada224:
In my check, segmentation fault in restarting did not occur after both of this fix and fix of #235. Even though @uemoto1 san is thinking that there is still problem in restarting, at least the issue that I pointed out last time was fixed. I'll merge.

Issue #232, opened 2017-11-13
Failure in RT mode of the multi-scale calculation


shunsuke-sato:
After PR #213 , the RT mode of the multi-scale calculation does not reproduce the original results.
@uemoto1 san, could you fix it?

uemoto1:
Thank you @shunsuke-sato san for bug reporting, I will check the reproduction of outputs !
And, Let me know the detail of the calculation condition and the related files?

shunsuke-sato:
As I wrote, the problem appears for the RT mode of the multi-scale calculation. So, if you set `calc_mode = 'RT'` and `use_ms_maxwell = 'y'`, you can see the problem.
If you do not see any problem, please let me know.

yhirokawa-ccs:
As related problem (probably), I found the application crashed when executing `example/bulk_Si/input_sc_Si.inp` with 16 nodes KNL after PR #213.
I may take a long time for finding the cause of problem, because I stay in Denver until next saturday to attend the conference.

yhirokawa-ccs:
Sorry, I missed the execution environment.
In above, I modified the input parameters as follows:

+ `num_rgrid = 16,16,16`
+ `num_kgrid = 24,24,24`

uemoto1:
@yhirokawa san. May I know the other parameters of the input files?  

I suspect the problem in `example/bulk_Si/input_sc_Si.inp` seems to be a exceeding of the memory allocation. Recently, we turned off the symmetry option from the example input files, in order to provide the generality as the templates. I think present `examples` requires more than 8 times larger memory and cpu resources than previous ones...

yhirokawa-ccs:
@uemoto1 san, no, I modified `num_rgrid` and `num_kgrid` only.
I hope the symmetry option sets to `Sym = 1` as the default, right?
For old codes (checked git revision: c302577a41958e36388a5305fc544ef09a8e4ed5), this parameter can executes under the same environment.

I calculate the wave function array size:

```
number of MPI processes = # of nodes = 16
number of kgrids = NK = 24^3
number of bands = NB = 32
number of rgrids = NL = 4096
wave function size in GS part: NK / 16 * 32 * NL = 1.6875 GB / node
wave function size in RT part: NK / 16 * (32/2) * NL = 0.84375 GB / node
```

I use MCDRAM only (16 GB / node), probably, this configuration not exceeds the memory capacity.

uemoto1:
[...]

PR #236, opened 2017-11-16
Fixed a bug of communicator and domain distribution.

yhirokawa-ccs:
This problem relates and discusses at #232.
In SC mode, `comm_*_tdks` aliases `comm_*_global`.


uemoto1:
Thanks @yhirokawa san, I will check it.

uemoto1:
Thank you for contribution, @yhirokawa san!!
 I confirmed the separated calculation of GS and RT is successfully finished!  
Presently the reproduction problem #232 still remains, But I think this bugfix is necessary to merge.

PR #212, opened 2017-10-28
Fixed a bug in the restart calculation

shunsuke-sato:
Prevent the initialization of current and vector potential in the restart calculation.

PR #237, opened 2017-11-16
Bugfix of RT calculation in MS mode.

uemoto1:
This pull request contains the bugfix claimed on the issue #232 .

### DIFF
--- a/src/ARTED/control/control_ms.f90
+++ b/src/ARTED/control/control_ms.f90
@@ -63,14 +63,12 @@ subroutine tddft_maxwell_ms
     end if
 
 !====RT calculation============================
-    if ( .not. (restart_option == 'restart')) then
-      if (trim(FDTDdim) == '2DC') then
-        ! TODO: FIx the initialization routine
-        call init_Ac_ms_2dc()
-      else
-        call init_Ac_ms
-      endif
-    end if
+    if (trim(FDTDdim) == '2DC') then
+      ! TODO: FIx the initialization routine
+      call init_Ac_ms_2dc()
+    else
+      call init_Ac_ms
+    endif
 
     rho_gs(:)=rho(:)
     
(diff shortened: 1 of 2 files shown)


src/ARTED/control/control_ms.f90
```diff
@@ -63,14 +63,12 @@ subroutine tddft_maxwell_ms
     end if
 
 !====RT calculation============================
-    if ( .not. (restart_option == 'restart')) then
```
shunsuke-sato, line 66: Why do you remove this `if` statement? I added it to avoid incorrect initialization in PR #212.

shunsuke-sato (commented):
@uemoto1 san, thank you for the implementation. I have a question. Please see below.

src/ARTED/control/control_ms.f90
uemoto1: @shunsuke-sato san, it is not directly affect to this bugfix. Which bug tracking, I have exhaustively tested  if statements related to restarting. Then, I found Line:66 is doubly checks same condition defined in Line:47. 

It has not the origin of Issue #237, therefore I think both existing and removing are acceptable in this code.

uemoto1:
@shunsuke-sato san, it is not directly affect to this bugfix. While bug tracking, I have exhaustively tested if statements related to restarting. Then, I found Line:66 is doubly checks same condition defined in Line:47.

It has not the origin of Issue #232, therefore I think both existing and removing are acceptable in this code.

uemoto1:
@ayamada224 san, Thank you for merging.
I have struggled at the printf debugging, and I found something is strange in `read_external_input_for_restart`. In the case `input_for_restart` is not exist, only the root process skips the reading and broadcasting, but any other processes can't skip the broadcasts. I think this mismatch leads the desync of MPIs...

uemoto1:
In this time, I have self-merge this PR, due to catch up the deadline of the next release...

PR #240, opened 2017-11-18
modify lmax and lloc for H atom in the C2H2 example

mn2007:
When Prof. Yabana generated H_rps.dat, he wrote a column of l=1 in the file because he intended to calculate nonlocal term. So I modify input files to follow Prof. Yabana's idea.


mn2007:
Oscillator Strength (z-direction, 9.28eV) calculated by jenkins is `1.208188e+00`, while it on k-computer is `1.216666e+00`. I'll consider how to fix this problem.

mn2007:
I think it's difficult to solve this problem in two days, so I decline this PR. I might give this PR again.

PR #244, opened 2017-11-21
Add opt_variables re-initialization routine.

yhirokawa-ccs:


uemoto1 (approved):
Currently, this modification has no side-effect to the main codes,so I think there is no problem to merging.

PR #246, opened 2017-11-23
fixed access violation of PS variables in MD option

ayamada224:
I fixed access violation of recently introduced PS relevant variables which occurs only in MD option (i.e. when PS potential is updated in time evolve simulation).


yhirokawa-ccs (approved):
It seems good. I'll merge it soon.

Issue #202, opened 2017-10-10
Slow convergence of force calculation


shunsuke-sato:
I found that SALMON shows very slow convergence for the grid spacing in the force calculation.
I checked the force in Silicon by adding a shift to one Si atom in the cell. The shift is 1% of the lattice constant (see the attached input for the details).

Figure below shows the force acting on the atom as a function of grid spacing. Here, the result with Octopus is also shown. One can clearly see that the SALMON has a severe problem in the force calculation even in the easiest material, Si, with the small distortion.

I guess that the problem originates from the issue #40 and/or #41 .

[input_sc_Si.inp.txt](https://github.com/SALMON-TDDFT/SALMON/files/1372368/input_sc_Si.inp.txt)

![force_salmon_oct](https://user-images.githubusercontent.com/6521050/31393022-03877114-addb-11e7-9a37-6bb4e948ae8e.png)

shunsuke-sato:
By PR #250, @ayamada224 san kindly fixed this problem, by implementing the wave-function gradient (see Issue #40 ).

I tested the force again and obtained the following result. Now, SALMON provides the consistent result with Octopus.

Then, I'll close this issue.

![force](https://user-images.githubusercontent.com/6521050/33504989-1057cd34-d6ea-11e7-9afc-6282c6a86f88.jpg)

PR #250, opened 2017-11-29
Improve force arted

ayamada224:
Force on ions in ARTED side were improved: 
Following @shunsuke-sato san's suggestions given in #202 , calculation method of force was changed to be a way using gradient of wave-function instead of gradient of potential. As a result, convergence of force for spacing was much improved.  The previous force calculation routines were kept: it can be switched by flag_use_grad_wf_on_force=.false./.true.. 

Some parts relating to the recent changes of pseudopotential variables were fixed for opt and md options.

The other tiny changes were done, too.


shunsuke-sato (approved):
@ayamada224 san, thank you for your implementation. I have checked this patch, and confirmed the improvement of the accuracy of force.

Then, I'll merge it.

PR #251, opened 2017-12-01
Add input variables lists in git repository

uemoto1:
As discussed in the previous meeting held on 22-24, Oct, we will prepare an input variables information, which is directory corresponds to the `Input Variables` and `Install and Run` in our wiki. The added markdown files `documents/*.md` are converted from our website http:/salmon-tddft.jp/.

TODO:
- Unify to the old (previous `manual/input_variables.md` file) into the new documents.

```
testmode=skip
```

### DIFF
--- a/documents/input_variables.md
+++ b/documents/input_variables.md
@@ -1,0 +1,442 @@
+We here summarize namelists that appear in this Tutorial. A thorough list of the namelist variables may be found in the downloaded file in 'SALMON/manual/input_variables.md'.
+
+# &units
+
+Mandatory: none
+
+```
+&units
+  unit_system='A_eV_fs'
+/
+```
+
+This namelist specifies the unit system to be used in the input file. Options are 'A_eV_fs' for Angstrom, eV, and fs, and 'a.u.' or 'au' for atomic units. If you do not specify it, atomic unit will be used as default.
+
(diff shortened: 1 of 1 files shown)


mn2007:
Thank you for preparing md files, @uemoto1 san. Could you give me some time to think about this PR? I wonder explanations do not cover whole variables implemented on SALMON. Maybe I'll comment this evening.

uemoto1:
Thanks @mn2007 san. I agree this documentation is not complete set of input variables, and this file is a simple copy of the wiki of the website. Then, I keep the previous keyword list as `manual/input_variables.md`. 

In the preparation of this MD file,  I found original `manual/input_variables.md` also has many typos and out-dated informations. I think we need a long time to follow up the documents to the present state. Therefore, in this PR, I suggest use added file as the template of the new "input variables", and update/fix them later.  (We need help of @mn2007 san to write the description related to Isolated System) And, if also keeps the old `manual/input_variables` as the reference during the updates.

uemoto1:
Next commitment,, I will remove `manual/input_variables.md` file from this PR.

mn2007:
I understand the situation that old .md file is in `manual` directory and new .md file is in `document` directory. Considering that the old .md file is unified into the new .md file in near future, current situation causes no problem. So I merge this PR. Thank you so much, @uemoto1 san.

uemoto1:
@mn2007 san, Thank you for quick checking and merging. Have a nice day!

Issue #252, opened 2017-12-01
Slowdown of solid-state calculations


shunsuke-sato:
Comparing with the old version (0cbe051d666bcb0ff10c6f7568ce54e3e04f0600), the latest version (72275fcb44be9f0c8de74d548b282a7335902bc3) gets more than 10% slower for the time-propagation. Especially, the total energy calculation seems to become ten times slower. It might be serious slowdown in some other conditions, especially in massive parallelization.

Probably, some of commits just before the first release caused this problem.

[input_sc_Si.zip](https://github.com/SALMON-TDDFT/SALMON/files/1522912/input_sc_Si.zip)

shunsuke-sato:
The issue was solved by PR #298. So, I close this issue.

Issue #40, opened 2017-06-02
Force calculation


shunsuke-sato:
I copied this issue from ARTED:
See: [Issue 1 in ARTED](https://github.com/ARTED/ARTED/issues/12)

In the pull request [PR8 in ARTED ](https://github.com/ARTED/ARTED/pull/8) I I explicitly excluded a small-radius component of spatial derivatives of pseudopotential projectors to stabilize the division. This treatment should be temporal, and we should find a good way for the spatial derivatives of pseudopotential projectors.

Here, I propose two ways to fix the problem:
Solution 1; We would explicitly prepare the value duVr(l)/r on the radial coordinate, and then use it instead of duVr(l) to construct the 3D projector. By this way, we can avoid the division by the small radius, r, in the 3D space.

Solution 2; Inner products of wavefunctions and gradient of pseudopotentials, <psi| grad pp>, can be replaced by inner products of pseudopotential and gradient of wavefunctions, <grad psi| pp>. By using this trick, we could quit using the spatial derivatives of the pseudopotential projectors in the force calculation.

Solution 1 is simpler for coding than Solution 2. However, Solution 2 would be more accurate if we employ Fourier filtering for pseudopotentials since the spatial derivatives enhance high wave number components of pseudopotentials.

Which way should we implement in ARTED? Or, are there any suggestions?

shunsuke-sato:
By PR #250, @ayamada224 san kindly implemented the wavefunction gradient for the force calculation. Then, this issue was solved. So, I'll close this issue.

PR #253, opened 2017-12-01
[WIP] Reduce calculation frequency of total energy

uemoto1:
This is PR contains a modification to speed up the total energy calculation claimed on issue #252.
At v.1.0.0, the total energy calculation is performed on every time steps, and now its frequency  `Total_Energy_omp` is reduced to 1/10 which is same to the previous implementation.

result

| Version | Total [sec] | Energy [sec]  |
| ------------- |:-------------:| -----:|
| This PR |  66.32 | 0.24 |
| v.1.0.0  | 70.96 | 2.18 |
| Before v.1.0.0  | 69.72 | 0.27 |

In the figure and tables above, the calculation result and computation time of the older implementation is reproduced. `SALMON/examples/Si_bulk_bench/input_sc.inp` is employed for this benchmark.

### DIFF
--- a/src/ARTED/control/control_sc.f90
+++ b/src/ARTED/control/control_sc.f90
@@ -186,14 +186,14 @@ subroutine tddft_sc
     if (use_ehrenfest_md == 'y') then
 !$acc update self(zu_t)
       call Ion_Force_omp(Rion_update_rt,calc_mode_rt)
-      ! if (iter/10*10 == iter) then
+      if (iter/10*10 == iter) then
         call Total_Energy_omp(Rion_update_rt,calc_mode_rt)
-      ! end if
+      end if
     else
 !$acc update self(zu_t)
-      call Total_Energy_omp(Rion_update_rt,calc_mode_rt)
+      ! call Total_Energy_omp(Rion_update_rt,calc_mode_rt)
       if (iter/10*10 == iter) then
-        ! call Total_Energy_omp(Rion_update_rt,calc_mode_rt)
+        call Total_Energy_omp(Rion_update_rt,calc_mode_rt)
         call Ion_Force_omp(Rion_update_rt,calc_mode_rt)
       end if
     end if
@@ -206,7 +206,7 @@ subroutine tddft_sc
     E_tot(iter,:)=-(Ac_tot(iter+1,:)-Ac_tot(iter-1,:))/(2*dt)
 
     Eelemag=aLxyz*sum(E_tot(iter,:)**2)/(8.d0*Pi)
-    Eall=Eall+Eelemag
+    ! Eall=Eall+Eelemag
     do ia=1,NI
       force_ion(:,ia)=Zps(Kion(ia))*E_tot(iter,:)
     enddo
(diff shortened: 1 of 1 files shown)


yhirokawa-ccs (approved):
I confirmed it, but I don't know the fix has any problem to numerical algorithm. I request a review/merge to other developer.

uemoto1:
@yhirokawa san, Thank you for reviewing and commenting. I agree this is a bit complicated decision , and I will continue to wait for reviewers.

---

For convenience to the other reviews, I have summarized the point. This problem is caused by the frequency increase of the total energy calculation which had been performed every one-tenth steps in the original v.0.0.1 series.  
 
The solutions which I have considered is as below:

1. Keep the present sampling rate of the output file, but reduce the total energy calculation by performing the linear interpolation to the total energy. `Proposed in This Pull Request`
  - **Merit**: Good Computation Speed and Sampling Rate
  - **Demerit**: Precision is not good as v.1.0.0, (but it is same as v.0.0.1)

2. Keep the present sampling rate and frequency of total energy calculation `Present Implementation of v.1.0.0`
 - **Merit**:  Good Sampling Rate and Precision
 - **Demerit**: Slow Computation Speed

3.  Reduce output sampling rate and frequency of energy calculation.
 - **Merit**: Good Computation Speed
 - **Demerit**: Low Sampling Rate of `SYSNAME_rt.data` file and lower the precision.

In this PR, I consider solution 1 is the most preferable and harmless implementation. The precision of the total energy is not worse than the v.0.0.1 and the other physical quantities (current/electric field...) are still keeping high sampling rate.

shunsuke-sato:
@uemoto1 san, thank you for having a look at the issue #40 .
Then, I would recommend the fourth solution, which is the original implementation:
Add a new file, and output the total energy every ten time steps in the new file instead of `SYSNAME_rt.data` file with.

I would say that the fine sampling is not needed for the total energy in most cases. Thus, we should not harm the speed by computing it every time step. On the other hand, the current requires the fine sampling for accurate analysis in most cases. Thus, we should not decrease the sampling rate all the quantities in `SYSNAME_rt.data`.
In this sense, this PR may be acceptable. However, I personally think that we should not output incorrect (or in accurate) values. Otherwise, the incorrect value may be used some serious analysis by users. Thus, I recommend to add a new file to avoid the above problems. Indeed, that was the original implementation.

However, you might have already had the discussion about these points in order to output all the quantities in `SYSNAME_rt.data`. So, please have a discussion with people in Tsukuba again.
Then, let us know the decision.

uemoto1:
@shunsuke-sato san, Thanks for the comments. I have discussed this problem with @yabana san and @mn2007 san. a few solution is raised.
1. Separate the total energy values from the present `SYSNAME_rt.data` output file. (This is same point suggested in @shunsuke-sato san's solution 4.) And provide some method to control the computation frequency of `total_energy` and `force` from user.

2. Speed up of the total energy calculation in periodic system. 

The formar will be the tempolaral solution and later will be resolved in future plan.

shunsuke-sato:
@uemoto1 san, thank you for discussing this problem. I agree the conclusion.

PR #254, opened 2017-12-02
Improved the performance of new force calculation.

yhirokawa-ccs:
I improved the performance of force calculation, which is implemented at PR #250.
It gets approximately 7.4x performance of currently implementation at RT/GS part.
The performance comparison uses 4 KNL nodes and `Si_gs_rt_pulse.inp`.

### DIFF
--- a/src/ARTED/common/ion_force.f90
+++ b/src/ARTED/common/ion_force.f90
@@ -54,8 +54,6 @@ contains
     real(8)      :: ftmp_l(3,NI),ftmp_l_kl(3,NI,NK_s:NK_e)
     real(8)      :: FionR(3,NI), FionG(3,NI),nabt_wrk(4,3)
     complex(8)   :: uVpsi,duVpsi(3)
-    complex(8)   :: zutmp0_3D(0:NLx-1,0:NLy-1,0:NLz-1)
-    complex(8)   :: dzudr0_3D(3,0:NLx-1,0:NLy-1,0:NLz-1)
     complex(8)   :: dzudr(3,NL,NB,NK_s:NK_e),dzudrzu(3),dzuekrdr(3)
 
     !flag_use_grad_wf_on_force is given in Gloval_Variable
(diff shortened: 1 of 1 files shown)


ayamada224:
Thanks @yhirokawa san, I'll check it today or tomorrow.

ayamada224:
I tested the improved code with Si system. I confirmed the speed of force calculation is much faster than the previous code without any changes of results, thanks!  Next I tested with H2 system, however, I found that it failed to calculate showing something MPI error. (There was no error if using the latest code, so I think this error is due to this PR change). I attached the input and output files. Could you please fix?

[H2_force_test.tar.gz](https://github.com/SALMON-TDDFT/SALMON/files/1524400/H2_force_test.tar.gz)

yhirokawa-ccs:
Thank you for checking and reporting. I found the failure of thread local memory (stack memory of a each thread) capacity.

The MPI error was caused by segmentation fault by `the stack memory limit exceeded`.  For fixing the bug, I modified that the allocation placement of temporal variables (`zutmp0_3D` and `dzudr0_3D`) to the global memory. At currently PR, I checked the H2 system calculation is successfully finished.

ayamada224:
Hi, @yhirokawa san, thanks for fixing. I tested again and I confirmed both of Si and H2 systems were properly calculated. 

Regarding calculation speed, the calculation time of Ion_Force are:
Si system: (16nodes/OFP)
17.7 sec (before improving(original))
2.9 sec (after first improving)
4.6 sec (after second improving)

H2 system: (1 node/OFP)
7.3 sec (before improving(original))
(failed) (after first improving)
11.6 sec (after second improving)

In Si system, the calculation time of fixed code is still better than the original. In H2 system, however, the calculation time after improving is slower than that of original. Maybe this is due to very small system? (# of electrons/orbitals is only 2) (but not sure). Even such a result, still there is benefit because calculation is faster for heavy system. Is further improvement possible? Even if not, I think it's OK, but I'm glad if you take a look at this point again. Thanks.

yhirokawa-ccs:
Thanks @ayamada224 san. Probably, the array dynamic allocation becomes considerable cost.
However, I realized that these temporal variables are not required in this calculation, so I removed it.

I checked the H2 system as same your environment:
- H2: 8.16 [sec] to 3.71 [sec]

yhirokawa-ccs:
Sorry, I closed this PR by mistake...
I reopened it.

ayamada224:
I tested. I confirmed the calculation speed is fast, but the obtained force was wrong....

H2 system: 
result by original:
 (forces on atoms)
       1      -0.000000       0.000000       0.082611
       2       0.000000      -0.000000      -0.082612

result by the latest:
(forces on atoms)
       1      -0.620108       0.000000       0.559617
       2       0.620107      -0.000000      -0.559617

yhirokawa-ccs:
I fixed the calculation error which is caused by a memory access bug of the stencil computation.

Except for the stencil computation, ARTED part is written by assuming the memory order is a form of (z,y,x).
However, the original implementation accesses with a form of (x,y,z), so it causes the calculation error...
I modified the memory access patterns in the stencil.

Result of H2 system:
```
 (orbital eigen energies of 1th k-point)
  1   -0.416329    2    0.047776
 (forces on atoms)
       1      -0.000000       0.000000       0.082611
       2       0.000000      -0.000000      -0.082612
```

ayamada224:
I see, I got it. I confirmed calculations by the fixed code give correct results with much faster calculation than original (both of H2 and Si systems). I'll merge.  
Thank you so much, @yhirokawa san.

PR #255, opened 2017-12-05
geometry optimization(ARTED) was replaced by force-driven CG method

ayamada224:
Geometry optimization (ARTED side) was replaced into force-driven CG method instead of finding energy minimum. This is because force calculation was recently improved by using wave function gradient (instead of potential gradient). Then, according to OCTOPUS paper, optimization should be done based on force only (not energy) as energy and force are now not consistent completely each other (total energy is not smooth due to egg box effect but force is smooth -- actually I suffered that optimization based on energy did not work well after improving force calculation)

I attached a test input file (one water molecule system) in this message.

Convergence of optimization is judged with maximum force. The default threshold is 1.0e-3 [au]. 

You can check iteration step in standard log file like
%> grep  Fmax  out.log

[H2O_opt_test.tar.gz](https://github.com/SALMON-TDDFT/SALMON/files/1529937/H2O_opt_test.tar.gz)

### DIFF
--- a/manual/input_variables.md
+++ b/manual/input_variables.md
@@ -966,11 +966,13 @@ Default is <code>0.5</code>.
 
 ## &opt
 <dl>
+<!--
 <dt>cg_alpha_ini; <code>Real(8)</code>; 3d</dt>
 <dd>
-Parameter for initial step length in line search in conjugated gradient method.
+Parameter for initial step length in line search in conjugated gradient method. (currently not available)
 Default is <code>0.8</code>.
 </dd>
+-->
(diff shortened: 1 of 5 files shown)


shunsuke-sato:
@ayamada224 san, thank you for your development. I tried to review, but I faced a deadlock. I need a bit more time to review it. Probably, I can finish the review this weekend. Sorry for inconvenience.

shunsuke-sato:
@ayamada224 san, thank you for implementing this function. I checked the structure optimization for bulk silicon, and it works well. Then, I'll merge.

shunsuke-sato (approved):

PR #256, opened 2017-12-07
Print rt wfn k

ayamada224:
I added a function of restarting RT (ARTED side) just using wave function (zu_t) but not using traditional option "restart_option" which print out all of huge data on memory. The new restarting way just prints minimal data and allow to change input keywords in the restart run if necessary.
The changes are followings:

(1)  wave function (zu_t) was printed out in "rt_wfn_k" directory after RT finished
(more exactly, occ, Rion, velocity are also printed in the directory)

(2) "read_rt_wfn_k=y/n" keyword was added (in &control).  
( then, similar keyword "read_initial_guess" was changed to "read_gs_wfn_k" )

How to restart RT:
(i) run with "GS" to get "gs_wfn_k" directory
(ii) run with "RT" (then use "gs_wfn_k")
(iii) Finaly, run with "RT" with "read_rt_wfn_k=y" option. (then must use "rt_wfn_k" and also "gs_wfn_k")

Please check that total energy is continuous. 
(but external field can not be taken over in restart run. This option does not work in multi-scale calculation)

### DIFF
--- a/manual/input_variables.md
+++ b/manual/input_variables.md
@@ -96,16 +96,22 @@ Default is the current directoy, <code>./</code>.
 <dd>Name of a filename for the restart calculation.
 </dd>
 
-<dt>read_initial_guess; <code>Character</code>; 3d</dt>
+<dt>read_gs_wfn_k; <code>Character</code>; 3d</dt>
 <dd>
-Option to read wave function as initial guess (pre-calculated "gs_wfn_k" directory) if this is <code>y</code>.
+Read ground state wave function as initial guess (pre-calculated "gs_wfn_k" directory printed by <code>calc_mode=GS</code>) if this is <code>y</code>.
 Default is <code>n</code>.
 </dd>
 
-<dt>modify_initial_guess; <code>Character</code>; 3d</dt>
+<dt>modify_gs_wfn_k; <code>Character</code>; 3d</dt>
 <dd>
-Option to modify initial guess wave function (pre-calculated "gs_wfn_k" directory) in combination with <code>read_initial_guess = y</code>.
+Option to modify initial guess wave function (pre-calculated "gs_wfn_k" directory) in combination with <code>read_gs_wfn_k = y</code>.
 If <code>copy_1stk_to_all</code> is set, the first k-point data file, wfn_gs_k0000001.wfn (supposed to be obtained by gamma-point calculation), is copied to all other points.
+Default is <code>n</code>.
+</dd>
+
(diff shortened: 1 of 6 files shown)


uemoto1:
Thanks @ayamada224 san, I will check it later.

uemoto1:
@ayamada224 san. I found a problem of the original backup does not works.
The SIGSEGV is happened when the program loads the ?????_re.dat file.
https://gist.github.com/uemoto1/96dc52ba512c38825377efcc6b1d5b70
```
##############################################################################
# SALMON: Scalable Ab-initio Light-Matter simulator for Optics and Nanoscience
#
#                             Version 1.0.0                                   
#
#   [Git revision] 016f1bc88e18a30e6a5ed1cffc69aa448066c3e5 in print_rt_wfn_k
##############################################################################
 Preprocessor: 
   ARTED_STENCIL_OPTIMIZED
   ARTED_STENCIL_PADDING
   ARTED_EXPLICIT_VECTORIZATION
   ARTED_ENABLE_SOFTWARE_PREFETCH
 parallel = Hybrid
 NUMBER_THREADS =          272
 Restore...  1.406669616699219E-005 sec =  3.907415601942275E-009 hour
 Restore time =  4.944300651550293E-002  sec
 This is the end of preparation for Real time calculation
 elapse time=   3.61031627655029      sec =  1.002865632375081E-003 hour
 -----------------------------------------------------------
forrtl: 致命的なエラー (174): SIGSEGV、segmentation fault occurred
```

uemoto1:
@ayamada224 san, can we have any input option to output (or surpress) `rt_wfn_k` because in some case the wavefunction data will be very large binary data. Therefore, I think it is happy for user to provide the cotrol method to create `rt_wfn_k` or not.

ayamada224:
Thanks, @uemoto1 san. To be able to control to write rt_wfn_k directory, I added option of "write_rt_wfn_k" in &control. Only if this keyword is y, rt_wfn_k directory is created. Similarly, I also added "write_gs_wfn_k". You can force to write GS wave function in "gs_wfn_k" in GS_RT calculation. (For "GS" calculation, gs_wfn_k is always written as before)

uemoto1:
Thank you @ayamada224 san. This modification will be very useful! I think it is worth to implement to the code. And I found the claimed problem above is not originating from this PR. I will write an issue about this.

I will continue reviewing...

uemoto1 (approved):

PR #258, opened 2017-12-11
fixed bug in PSmask

ayamada224:
A bug in PSmask option was fixed: 
Previously, extremely large energy or NaN showed up when KY pseudopotential of H atom was used (division by zero or very small number was done depending on pseudopotential data). This was fixed.


uemoto1:
Thank you @ayamada224 san, I will check it later.

PR #259, opened 2017-12-13
Add deallocation of IPIV and WORK

uemoto1:
This PR is bugfix of memory leak in initialization of pseudopotential data.


shunsuke-sato:
@uemoto1 san, thank you for your development.
I reviewed it. But, if I understand correctly, these local arrays are automatically deallocated on exit by the Fortran standard.

Or, should we explicitly deallocate local arrays in every routines?

uemoto1:
@shunsuke-sato san, I checked it and I found it is my missunderstandings. I withdraw this PR. Thank you for reviewing!

PR #262, opened 2017-12-23
improve allocate for large system to avoid crash

ayamada224:
Allocation of two variables for pseudopotential in hpsi.f90 was improved: When large system with small spacing is calculated, the calculation crashes even though (probably) the used memory size is smaller than the memory of the node machine. It occurs around the variables of "psendo" and "dpseudo" in subroutine "pseudo_pt", and it seems due to something their memory or allocation. I empirically fix such a case by using allocate sentence instead of declaration sentence.

### DIFF
--- a/src/ARTED/common/hpsi.f90
+++ b/src/ARTED/common/hpsi.f90
@@ -162,10 +162,9 @@ contains
 
       integer    :: ia,i,j,ip,ioffset
       complex(8) :: uVpsi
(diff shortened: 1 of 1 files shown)


yhirokawa-ccs:
@ayamada224 san, I apologize for late reply.
Probably, dynamic memory allocation in a small size routine decreases the performance. I'll send a revised commit soon.

yhirokawa-ccs (approved):

PR #264, opened 2017-12-29
changed print of force

ayamada224:
I changed way of force printing (ARTED side) like:
(1) In RT mode, force is not printed by default (but printed in md option by default)
Then, force is not calculated reducing calculation time (not necessary).
(2) Force is printed in "xxxx_trj.xyz" file together with xyz and velocity (instead of xxx_force_dRion.out is not used). 
(3) Time step interval for force printing can be controlled by using new input variables in &analysis, like:
     out_rvf_rt="y", 
     out_rvf_rt_step=10

BTW, after latest maintenance of OFP, intel module has been updated and default version is intel/2018.1.163. If you use it for compiling SALMON, the force calculation does not work printing "infinity" or "*******". I don't know why nor how to solve. So, currently, please change to the previous module version like
module purge
module load intel/2017.4.196

Sorry for asking in winter vacation. You can review it next year of course. Thanks. Have a nice holiday!

### DIFF
--- a/src/ARTED/control/control_sc.f90
+++ b/src/ARTED/control/control_sc.f90
@@ -54,8 +54,6 @@ subroutine tddft_sc
   select case(use_ehrenfest_md)
   case('y')
     Rion_update_rt = rion_update_on
-    write(comment_line,110) 0, 0
-    call write_xyz(comment_line,"new","rv ")
   case('n')
     Rion_update_rt = rion_update_off
   end select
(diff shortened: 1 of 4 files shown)


uemoto1:
@ayamada224 san, Thank you for contribution! I will check it later.

uemoto1 (commented):
Thank you. I think it is available to merge the developer branch.

PR #267, opened 2018-01-08
Improve the memory allocation in pseudo-pt.

yhirokawa-ccs:
The working memories (`spseudo` and `dpseudo`) move to `opt_variables` module.
See also: PR #262


shunsuke-sato:
@yhirokawa san, thank you for your implementation. I have reviewed and confirmed it. So, I'll merge this patch.

PR #269, opened 2018-01-08
Improve the performance of OpenACC codes.

yhirokawa-ccs:
1. Apply OpenACC to a few CPU code.
1. Improve the performance a few OpenACC-implemented routines.
1. Reduce memory transfer costs between CPU and GPU.


shunsuke-sato:
@yhirokawa san, thank you for your implementation. However, I do not have a proper access to GPU machines. So, I could not check it.
I would like to ask @uemoto1 san to have a look at this PR.

uemoto1:
Sorry for late reply @yhirokawa san, since I have been dreadfully busy in recent a few days.  I will check your pull request as above.

yhirokawa-ccs:
@uemoto1 san, thank you for merging. I'll send a mini PR for wrapper scripts of GPU execution.

PR #272, opened 2018-01-15
Hotfix for Intel compiler 2018.

yhirokawa-ccs:
When compiling with Intel compiler version 2018, I found a bug of `ion_force` calculation which is caused by `omp reduction` in multi-dimensional array.
I replace `omp reduction` clause to `omp critical` section.

### DIFF
--- a/src/ARTED/common/ion_force.f90
+++ b/src/ARTED/common/ion_force.f90
@@ -67,7 +67,7 @@ contains
 
       !ion-ion: Ewald short interaction term (real space)
       ftmp_l=0.d0
-!$omp parallel do private(ia, ik,ix,iy,iz,ib,rab,rab2) reduction(+:ftmp_l) collapse(5)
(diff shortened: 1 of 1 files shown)


ayamada224:
Thanks for that bug fixing, @yhirokawa san. I have been suffered from the bug by intel compiler ver. 2018 but didn't know how to fix. I'll review it as soon as possible.

ayamada224:
I tested it with Si and H2 systems, however, forces printed during GS calculation are still bad like "********". I calculated on OFP and used modules are like:

$>  module list
Currently Loaded Modulefiles:
  1) impi/2018.1.163    2) intel/2018.1.163

If you need I'll put the input files.

yhirokawa-ccs:
@ayamada224 san, sorry I fixed it because the patch is defective.
I tested it with Si system rest on `testsuites` directory. I used OFP with module `intel/2018.1.163` and `impi/2018.1.163`.

PR #273, opened 2018-01-17
Add manual reduction routine for OpenMP in `ion_force` subroutine.

yhirokawa-ccs:
I apologize that the GS part calculation takes a long time by applying PR #272.
This problem is caused from `omp critical` section that has high overhead.

I implemented parallel reduction routine for OpenMP, which is based on tree-based reduction algorithm.
I checked `forces on atoms` output with the Si and H2 system.


ayamada224:
Thanks, @yhirokawa san, I confirmed the calculation results did not change but calculation time of Ion_Force went back to normal. I merge it.

PR #278, opened 2018-01-24
Speed-up update of pseudopotential in md option

ayamada224:
Update of pseudopotential was speeded up in MD option (ARTED side). (subroutine prep_ps_periodic), as  the updates of pseudopotential are very often in MD option, which is bottle neck of calculation speed. One keyword for that step_update_ps2 was also added.

Please just check if the results of GS, RT (w/o md option) are the same as previous version.


yhirokawa-ccs (approved):
It looks like there are no problems. I'll merge.

ayamada224:
Thanks, @yhirokawa san!

PR #281, opened 2018-02-02
Add analysis options: prints of transition moment, transition difference density, no-update of Hamiltonian(functional)

ayamada224:
Three useful analysis options are implemented in ARTED side:

(1) Print transition moment between orbitals
Giving a new keyword, out_tm='y'  in &analysis block in 'GS' mode, transition moment <u_nk|p|u_mk> and relating values for non-local pseudopotential between Bloch orbitals are printed in SYSname_tm.data. It can be useful information as oscillator strength and absorption spectra (or conductivity and dielectric function) etc can be obtained from these values in the ground state. 

(2) Print cube/vtk files of transition difference density
Giving new keywords, out_dns_trans='y' and out_dns_trnas_energy=xxx in &analysis block in 'RT' mode (assuming use of Impulse field), different density involving transition by oscillating field is calculated, which is calculated as FT(rho(r,t)-rho_gs(rt)) at the user specified energy/omega value. Then, out_dns_trans_Re.cube/vtk and out_dns_trans_Im.cube/vtk are printed. This is useful to know what's happened in density at specific energy (especially at peak of absorption spectrum)

(3) Option to use ground state Hamiltonian in RT mode without updating functional during time evolution.
Giving a new keyword, no_update_func='y' in &functional block, Hamiltonian (more specifically Hartree & ExCor) is not updated during time-evolution in RT mode (i.e. use of ground state Hamiltonian), which is just for particular purpose to know how much important the time-dependent of Hamiltonian is.

### DIFF
--- a/manual/input_variables.md
+++ b/manual/input_variables.md
@@ -880,16 +880,16 @@ If <code>'y'</code>, density is output.
 Default is <code>'n'</code>.
 </dd>
 
-<dt>out_elf; <code>Character</code>; 0d</dt>
-<dd>
-If <code>'y'</code>, electron localization function is output.
-Default is <code>'n'</code>.
-</dd>
-
 <dt>out_dns_rt/out_dns_rt_step; <code>Character/Integer</code>; 0d/3d</dt>
 <dd>
 If <code>'y'</code>, density during real-time time-propagation is output
 every <code>outdns_rt_step</code> time steps.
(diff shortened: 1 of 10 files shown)


syamada0 (approved):

syamada0:
Thank you @ayamada224 san.
I checked it and could not find any problems.

PR #282, opened 2018-02-05
Bugfix of Electron Density Plot

uemoto1:
This pull request provides the bugfix in the export of the electron densities; the lack of the `comm_is_root` checks. Presently, this routine does not make any problems in the calculation, however we expect it will cause the significant I/O troubles in the future implementation.


uemoto1:
@yhirokawa san, Thank you!

PR #287, opened 2018-02-09
support restart of electric field using rt_wfn_k directory

ayamada224:
Recently so far, restarting of RT mode with minimum printing (in "rt_wfn_k" directory) instead of traditional "restart_option" has been implemented (ARTED side), but field had not been taken over after restarting. Now  this got to support the electric field, i.e. after restarting using "write_rt_wfn_k=y" & "read_rt_wfn_k=y", the field oscillation continues. (but currently only for ae_shape1=AcosX ). 
Relating that, "t1_delay" option was added (delay of the first pulse)

### DIFF
--- a/manual/input_variables.md
+++ b/manual/input_variables.md
@@ -117,7 +117,7 @@ Default is <code>n</code>.
 
 <dt>read_rt_wfn_k; <code>Character</code>; 3d</dt>
 <dd>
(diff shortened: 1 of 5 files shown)


uemoto1:
Thank you @ayamada224 san for developing, I will check it later...

uemoto1 (approved):
Thank you @ayamada224 san. I confirm your PR works well.

PR #291, opened 2018-02-20
fix a problem for rho_dne

mn2007:
In the conversation with @ayamada224 san, I found a mistake for a definition of rho_dne in the GCEED part. @ayamada224 san, could you check this PR?


mn2007:

PR #292, opened 2018-02-21
modify a discription of a manual

mn2007:
I noticed a mistake in a manual, so I fix it. @tt-ims san, could you check this PR?


mn2007:
Thank you for merging so quickly, @ayamada224 san.

PR #293, opened 2018-02-22
Add decomposition analysis option of projection: docompose the number of excited electron into atom

ayamada224:
A keyword "projection_decomp" was added. If "projection_decomp=atom" combined with "projection_option="gs",  decomposition of the number of excited electron into atoms is done and it's printed in SYSname_nex_atom.data.


uemoto1:
Thank you @ayamada224 san, I will check this PR.

PR #294, opened 2018-02-26
Fix bug in Ecos2 type incident field in ARTED & Made normalization of polarization vector

ayamada224:
A bug of recently implemented "Ecos2" option of incident electric field in ARTED side was fixed: The polarization vector epdir_re1 was missing (i.e. equivalent to use of (1,1,1) polarization vector ). 

Also I added normalization of polarization vector (epdir_reX, epdir_imX) after reading input file to avoid confusing.


uemoto1 (approved):
Thank you, @ayamada224 sam. I will confirm the PR is available to merge.

PR #295, opened 2018-03-16
Bug fix probe field in Ecos2 option (missing polarization vector...)

ayamada224:
Bug in Ecos2 probe field (in ARTED) was fixed (polarization vector was missing like the recent found bug of pump field of Ecos2 ....)


uemoto1 (approved):
Thank you @ayamada224 san for Bugfix.
I will merge it into the developer branch.

PR #296, opened 2018-03-20
amplitude is devided by omega in an Acos2 calculation

mn2007:
I found that an amplitude was not devided by omega in the GCEED part while it was in the ARTED part. So I fix. I also show a calculated result. 

rt


uemoto1:
Thank you @mn2007 san for developing, I will check this later...

mn2007:
Thank you in advance for checking, @uemoto1 san.

PR #298, opened 2018-04-16
Solved the slowdown problem due to the energy calculation

shunsuke-sato:
I solved the issue #252, which is the slowdown of the solid state calculation due to frequent computation of the total energy. 
In this patch, SALMON separately outputs the energy data in an additional file, `SYS_name_rt_energy.data`.
This solution was decided in the discussion of PR #253.

This patch recovers the computational performance: it makes more than 20% speed up, compared with the present `develop` branch.


uemoto1 (approved):
Thank you @shunsuke-sato san for developing, and I confirm this code works well.

shunsuke-sato:
@uemoto1 san, thank you for your quick review and merge.

PR #300, opened 2018-05-23
Add switches to surpress the radiation / reflection in multiscale calculation

uemoto1:
##Overview

This PR provides a parameter `debug_switch_no_radiation` for the check/test of multiscale calculation.
If this switch enabled in input file (or the number of macropoints is set to 0), the macroscopic system calculation neglects the current density by microscopic system. It is used for the calculation check of the radiation and reflection field.

This modification is only for a temporal implementation before merging the common FDTD module.


ayamada224:
I confirmed it works well, thanks!

ayamada224 (approved):
Approved

PR #304, opened 2018-06-08
Implement md option to multiscale calculation and some relating modifications

ayamada224:
MD option (use_ehrenfest_md=y) became available with multi-scale calculation.  
Then, new input keywords was implemented: 
set_ini_coor_ver=y/n
read_wfn_k_ms=y/n
which enable to read atomic coordinates, velocities, and ground state wave-function of each macro-grid point separately. 
For convenience to treat tons of input/output files, directory structure was changed: these all files are put in "multiscale" directory and the files for each macro-grid point are put in "MXXXXXX" (e.x. M000001/, M000002/ ; macro-grid point index) and files along time evolve are put in "RT_Ac".

### DIFF
--- a/manual/input_variables.md
+++ b/manual/input_variables.md
@@ -103,7 +103,7 @@ Default is the current directoy, <code>./</code>.
 
 ><dt>write_gs_wfn_k; <code>Character</code>; 3d</dt>
 ><dd>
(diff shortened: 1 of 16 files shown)


uemoto1:
Thank you, I will send the modified test routine...

PR #307, opened 2018-06-20
multi-scale + md and restart option

ayamada224:
Again, multiscale + md option. 
Some debugs are done.

Also, restart of multiscale by new keywords "write_rt_wfn_k_ms" and "read_rt_wfn_k_ms" is available, which is different from the all back-up option so far. This works for multi-scale calculation with and without md option.


uemoto1:
Thank you @ayamada224 san, I will check this PR.

PR #310, opened 2018-06-24
multi-scale + md and restart option

ayamada224:
Again, multiscale + md option.
Some debugs are done.

Also, restart of multiscale by new keywords "write_rt_wfn_k_ms" and "read_rt_wfn_k_ms" is available, which is different from the all back-up option so far. This works for multi-scale calculation with and without md option.


uemoto1 (commented):
Thank you @ayamada224 san, I guess the problem seems to be in the verification process of the test suite scripts. I will also check it.

uemoto1:
Thank you @ayamada224 san for developing. I confirmed this PR works well.

uemoto1 (approved):

PR #314, opened 2018-06-27
fix a bug and modify automatic assignment of processes in lsda calculations

mn2007:
It seems that bugs appear in real-time propagation for lsda calculations when orbital parallelization is used. So I modify the assignment of processes not to use the orbital parallelization in the lsda calculations. @tt-ims san, could you check?


tt-ims:
OK, I will check it soon.

tt-ims:
I approve this PR.

mn2007:

PR #305, opened 2018-06-19
Exchange Correlation Functionals

uemoto1:
## Overview

This pull request provides a new module of the exchange-correlation functionals (`salmon_xc`), which contains the ordinally ARTED functionals (e.g. PZ, PZM, PBE, and TBmBJ) and Libxc wrapper routines. The Libxc wrappers are enabled when the libxc is linked in the building process.
For the building, the new arguments of the configure script are added as below:
- `--enable-libxc`: link the Libxc library (by automatic detection in CMake)
- `--with-libxc=PATH/TO/LIBXC`: specify the Libxc prefix directory to link with to the SALMON.

### Build without Libxc
You can also build the SALMON by the ordinary installation procedure.

### Build with Libxc
If you want to use Libxc mode, you must do adittional preparation.
Download the source-list and install by the website(http://www.tddft.org/programs/libxc/installation/):
```
wget http://www.tddft.org/programs/octopus/down.php?file=libxc/4.2.1/libxc-4.2.1.tar.gz
tar -zxvf libxc-4.2.1.tar.gz
```
Enter to the libxc source directory:
```
./configure --prefix=PATH/TO/LIBXC
make; make install
```
Finally, you can execute configure script of SALMON with specifing the Libxc directory:
```
configure.py --arch=ARCHITECTURE --prefix=PREFIX --with-libxc=PATH/TO/LIBXC
make; make install
```
 
### New input parameters
You can specify the xc-functional by using the `xc` parameter:
```
&functional
  xc = 'pz'
/
```

| xc | Family | Detail |
| --- | --- | --- |
| pz | LDA | Built-in PZ LDA |
| pzm | LDA | Built-in PZM LDA |
| pbe | GGA | Built-in PBE GGA |
| tbmbj | MetaGGA | Built-in TBmBJ |
| bj_pw | MetaGGA | Built-in Becke-Johnson |
| libxc_pz | LDA | Libxc PZ LDA |
| libxc_pzm | LDA | Libxc PZM LDA |
| libxc_pbe | GGA | Libxc PBE GGA |

In addition, you can also specify the Libxc variables individually:
```
&functional
  alibx = 'LDA_X'
  alibc = 'LDA_C_PZ_MOD'
/
```
The  xc-functional list is available in the website (http://www.tddft.org/programs/libxc/functionals/).

PR #316, opened 2018-06-28
Remove non-used subroutines

uemoto1:
This pull request removes the non-used subroutine (functions) previously implemented in the exchange-correlation modules in ARTED part. The corresponding subroutines are already moved to common `xc` module by PR #305 .


mn2007:
Since all checks have passed, I merge this PR. Thank you, @uemoto1 san.

PR #318, opened 2018-06-29
fix a bug for the Broyden's method

mn2007:
Based on @ayamada224 san's report about out of array access, I fix a bug for the Broyden's method. @ayamada224 san and @uemoto1 san, could one of you two check this PR?


uemoto1:
Thank you @mn2007 san, I will check your PR.

uemoto1:
@mn2007 san, I confirm that the problem in the Broyden of GS calculation seems to be resolved.

mn2007:

PR #322, opened 2018-07-02
Fixed the hangs in Ion_Force subroutine.

yhirokawa-ccs:
In some situations, `Ion_Force` subroutine hangs because the `dzudr` variable requests huge space to the stack memory.
From now on, would you please check each array size on the stack to avoid memory allocation error.


uemoto1:
@yhirokawa san, thank you for your contribution. I will read your patch.

Presently, one of the regression test of `diana-gpu01` raise the error code,
due to the temporal problem in script to check the calculation results.
(This is not a critical problem of code, I guess it is possible to merge 
with ignoring the error from `diana-gpu`)

Since I'm preparing the fix of the GPU testcases, please wait for a moment.

Sorry for inconvenience.

uemoto1:
I confirm your PR works well, and it is possible to merge.
Thank you for your contribution!

PR #323, opened 2018-07-02
Modify output format of floating point number

uemoto1:
In this pull request, the output format of the floating point number is changed from `ES22*` to `E23*`. It solves the problem in PGI Fortran environment to write the real number which is much smaller than 1E-99.


mn2007:
Testsuites for the ARTED part successfully end in both knl and gpu. So I merge this PR.

uemoto1:
Thank you @mn2007 san for merging!

PR #327, opened 2018-07-07
modify testsuite

mn2007:
I found that it takes 25 minutes to calculate systems of testsuites for isolated systems in a gpu environment. This modification reduces calculation times to about 5 minutes.


tt-ims:
I approve this PR.

mn2007:

PR #328, opened 2018-07-24
check input for xc calculations

mn2007:
If users use bad combination for input, the run is terminated.


mn2007:
It seems that checks have passed. @tt-ims san, could you check this PR?

tt-ims:
I approve this PR.

mn2007:

PR #332, opened 2018-08-24
Modify CMake Platform files.

yhirokawa-ccs:
The software prefetch optimization will be disabled in `intel-knl` platform as a default.


uemoto1:
Thank you @yhirokawa san.
I confirmed your pull request works well.

PR #335, opened 2018-09-07
modify standard output about FFTE

mn2007:
Even in the situation that users do not use FFTE, starndard output about FFTE is written at the beginning for the GCEED part. This PR erase it. Of course, it remains when users use FFTE.


tt-ims:
I will check this PR soon.

tt-ims:
I approve this PR.

mn2007:

PR #337, opened 2018-09-08
modify OUT_IN_data.f90 not to make access outside bounds of arrays.

mn2007:
This PR does not affect the normal use of SALMON.

In the GCEED part, the access outside bounds of arrays occurs when users read checkpoint files in scf calculations with k points parallelization or orbital parallelization. This PR fix the bug for the access.


tt-ims:
I approve this PR.

mn2007:

PR #339, opened 2018-09-10
modify checks for convergence

mn2007:
Options for convergence other than rho_dne can be used when iperiodic=3 and domain_parallel = 'y'. This PR enables for uses to use the options.


mn2007:

PR #340, opened 2018-09-17
Fixed OpenMP reduction error.

yhirokawa-ccs:
In ARTED part, SALMON outputs `NaN` error in GS calculation (forces on atoms) when executing number of thread is not power of two. I fixed it.

### DIFF
--- a/src/ARTED/common/ion_force.f90
+++ b/src/ARTED/common/ion_force.f90
@@ -44,6 +44,7 @@ contains
     use timer
     use salmon_math
     use projector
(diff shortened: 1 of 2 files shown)


uemoto1:
Thank you @yhirokawa !
I will check your pull request.

uemoto1:
I understand this pull request contains the fix of the allocation and initialization problem of `ftmp_t` variables which cause the violation in the `mreduce_omp` subroutine. I agree this bugfix is necessary to merge. Thank you for the contribution.

PR #341, opened 2018-09-18
Fix initialization problem of rho_nlcc in ARTED part

uemoto1:
This pull request provides a bugfix in the exchange-correlation calculation in ARTED part. I found an initialization error of the NLCC (nonlocal core correction) charge densities. Due to this problem, the present version causes the NaN overflow when a certain pseudopotential (which uses NLCC) is employed.


mn2007:
I confirmed that a NLCC calculation for a dielectric function of cBN successfully finished by using 72 processes on FX10, so I merge this PR.

uemoto1:
Thank you @mn2007 san for checking and merging!

PR #342, opened 2018-09-19
Skip the atomic coordinates section in the input file for classical electromagnetic problem

uemoto1:
This pull request contains a small modification which skip to read `atom_(red_)?coor`  section in the input file if the number of atom `natom` is specified to 0.  This change makes possible to simplify the input file when we use the classical electromagnetic problem which does not use the atomic structure.


tt-ims:
@uemoto1 san,
thank you for your quick modification. I will check this PR soon.

tt-ims:
I approve this PR.

PR #345, opened 2018-09-24
Fix the Build on Single Process Environment

uemoto1:
## Overview

This PR contains a fix of the *dummy* MPI-communicator which currently does not work on the single process (no-MPI) environment. 

- Add a dummy-communicator `comm_alltoall` to follow up the recent modification. 
- Unify `salmon_communication.f90` and `salmon_communication_dummy.f90` for simplification of future developing.
- Add macro-variable `SALMON_USE_MPI` to switch the multi/single process code.

## Build on CMake
Use `--disable-mpi` option to build for *no-mpi* environment:
```
configure.py --arch=PLATFORM --prefix=PREFIX --disable-mpi 
make -j 4
make install
```

### Gnu make
For convenience, it is also possible to use `gnu_makefiles/Makefile.PLATFORM-without-mpi`.


mn2007:
Linear response calculation of a C2H2 molecule without MPI successfully finished on FX100. So I merge this PR. Thank you, @uemoto1 san.

uemoto1:
Thank you for checking, @mn2007 san!

PR #348, opened 2018-10-09
Fixed compile error due to fortran free/fixed format.

yhirokawa-ccs:
Fortran compiler detects the free/fixed format in each source file automatically.
But GCC `-ffree-form` option interprets that all source files are the free format even fortran77 file.

GNU Makefile for GCC happens compile error in FFTE extension (fortran77 file), I fixed it.
In addition, I fixed the compile error due to column length in the free form.


uemoto1:
I confirm your PR. @yhirokawa san, Thank you for contribution.

PR #349, opened 2018-10-11
Fixed manual thread reduction code in ion_force.f90

yhirokawa-ccs:
The `array index out of bounds` error happens in `ion_force.f90` when OMP_NUM_THREADS is not set to the power of two number, I fixed it.


yhirokawa-ccs:
I found error, please wait a check it.

yhirokawa-ccs:
OK, I fixed it.

mn2007:
I checked atomic forces for examples/bulk_Si with 8 processes on OFP with GCC compiler + OpenMPI. When I changed number of threads as 63, 64 and 65, atomic forces written in standard output file was completely same. So I merge this PR.

PR #360, opened 2018-10-24
modify files related to scalapack

mn2007:
This modification is first aid for doing cmake with scalapack option. Implementation of scalapack version for periodic systems in domain parallel is a future task. 

@yhirokawa san, is it possible to check this PR?


yhirokawa-ccs (approved):

mn2007:
Thank you for merging, @yhirokawa san.

PR #364, opened 2018-10-29
add function of gscg

mn2007:
In a CG routine, a method like modified Gram Schmidt is used. Although this method ensures the accuracy, it causes a bottleneck of communication. This PR enables for users to choose a method like normal Gram Schmidt. 

@tt-ims san, could you check this PR?


tt-ims:
OK, I will check it.

tt-ims:
I confirmed that gscg option normally works. I approve this PR.

mn2007:

PR #365, opened 2018-11-20
modify info file

mn2007:
Number of electrons written in info file was wrong. So, I modify.


tt-ims:
I confirmed that the info file is correctly generated. I approve this PR.

mn2007:
Thank you for checking and merging, @tt-ims san.

PR #368, opened 2018-11-26
Remove manual and document directories

ayamada224:
I removed manual and document directories because these are now managed in the separated SALMON-DOCS repository.


uemoto1:
Thank you for @ayamada224  san. I will merge this patch.

PR #373, opened 2018-12-19
organize code of some trial functions in multi-scale

ayamada224:
Just organize the code to be clean for some trial functions of multi-scale model (this does not affect the calculations)


uemoto1 (approved):
@ayamada224 san,  I confirmed the regression test has been passed and I will merge it.
Thank you for contribution!

PR #376, opened 2018-12-21
modify directory structure of examples

mn2007:
I modify the directory structure of examples related to domain parallel.


mn2007:

PR #377, opened 2018-12-21
Pseudopotential routines

uemoto1:
Move pseudopotential routines to src/atom/pp


mn2007:
I'm so surprised by this easiest solution. Thank you for implementing, @uemoto1 san.

mn2007 (approved):
I approve this PR.

PR #380, opened 2018-12-25
Add ehFDTD examples

tt-ims:
This PR adds examples for ehFDTD.


tt-ims:
Thank you for checking and merging, @ayamada224 san!

PR #386, opened 2019-01-16
move some variables from pp_info to pp_grid

mn2007:
According to the request from @syamada0 san, I move some variables from pp_info to pp_grid.


syamada0:
Thank you for modifying, @mn2007 san!

syamada0 (approved):

mn2007:
Thank you for merging so quickly, @syamada0 san.

PR #387, opened 2019-01-22
calculate jxyz in parallelized domain

mn2007:
I modify codes to calculate jxyz in parallelized domain. @syamada0 san, could you check?


syamada0 (approved):

mn2007:
Thank you for merging so quickly.

PR #388, opened 2019-01-22
shift jxyz

mn2007:
I shift jxyz in a unitied routine. This modification is applied to domain parallel version for systems with periodic boundary condition.


mn2007:
Somehow displayed number of changed files is 0. So I close this PR and make new branch.

PR #389, opened 2019-01-23
shift jxyz

mn2007:
I shift jxyz in a unitied routine. This modification is applied to domain parallel version for systems with periodic boundary condition.


syamada0 (approved):

mn2007:
Thank you for merging, @syamada0 san.

PR #393, opened 2019-02-25
Fix of version number printing

uemoto1:
This pull request contains the fix of version number printed in the stdout.


uemoto1:
Thank you @yhirokawa san

PR #394, opened 2019-03-18
Fixed invalid memory reference in ARTED/preparation/prep_ps.f90

yhirokawa-ccs:


mn2007:

PR #396, opened 2019-03-25
Fix error: array index out of bounds.

yhirokawa-ccs:


mn2007:

